PR #93, opened 2021-06-09
Allow replicas/size override for master/slaves

davidcollom:
Add ability to override the replica/size for master/slave, this is for a scenario where you may not want any slaves as your application is write heavy, or that your application is read heavy and you want 2x or 3x slaves over masters.


iamabhishek-dubey:
Hey @davidcollom ,
Once you are creating the replicas of slave or master, where do you have the logic to add them in the cluster? For example:- I have 3 masters and I want 6 slaves for it. How are you handling that situtation

iamabhishek-dubey:
Hey @davidcollom ,
can you please hold your PRs for 2-3 days, actually I am introducing a breaking change in which we are separating the redis standalone and cluster CRDs and making the code-based improvements

davidcollom:
Hey @iamabhishek-dubey Thanks for getting back to me, sorry some of these PR's are little inter-twined 😞  - I tried to split them out.

The logic still stands as before, 3x Master, 6x Slaves would mean 9 Replicas in total with https://github.com/OT-CONTAINER-KIT/redis-operator/pull/93/files#diff-b83693ad5ae2859ef2bc0b066d04fd577f9ed8291c3d46a82014ba10dfef847cR98-R106 we would then use these values to ensure each replica is ready.

Additionally #92 Will also improve this, as it won't try and create a new cluster, but only add any missing replicas, meaning slaves can be scaled Up/Down easier.

iamabhishek-dubey:
Please fix merge conflict

api/v1beta1/redis_types.go
```diff
@@ -26,21 +26,20 @@ import (
 
 // RedisSpec defines the desired state of Redis
 type RedisSpec struct {
-	KubernetesConfig  KubernetesConfig             `json:"kubernetesConfig"`
-	RedisExporter     *RedisExporter               `json:"redisExporter,omitempty"`
-	RedisConfig       map[string]string            `json:"redisConfig"`
-	Resources         *corev1.ResourceRequirements `json:"resources,omitempty"`
```
davidcollom, line 32: This was never referenced or used, as its within the `KubernetesConfig`

api/v1beta1/redis_types.go
```diff
@@ -26,21 +26,20 @@ import (
...
+	SecurityContext   *corev1.PodSecurityContext `json:"securityContext,omitempty"`
+	PriorityClassName string                     `json:"priorityClassName,omitempty"`
+	Affinity          *corev1.Affinity           `json:"affinity,omitempty"`
+	Tolerations       *[]corev1.Toleration       `json:"tolerations,omitempty"`
 }
 
 // RedisStatus defines the observed state of Redis
 type RedisStatus struct {
-	Redis RedisSpec `json:"redis,omitempty"`
+	Redis RedisSpec `json:"redis"`
```
davidcollom, line 42: This should be a required field, given that its critical..

config/crd/bases/redis.redis.opstreelabs.in_redisclusters.yaml
```diff
@@ -2322,19 +2336,6 @@ spec:
...
-      name: Master
-      type: integer
-    - jsonPath: .spec.clusterSize
-      description: Current slave node count
-      name: Slave
-      type: integer
-    - jsonPath: .metadata.creationTimestamp
-      description: Last Deployment Time
-      name: Last Deployment Time
-      type: date
```
davidcollom, line 2337: Not sure why this is being removed, nor where its being referenced/defined.

api/v1beta1/redis_types.go
iamabhishek-dubey: This is being used here

https://github.com/OT-CONTAINER-KIT/redis-operator/blob/716e52b07bdc44d808512b291d5e7da5b91d527d/k8sutils/redis-cluster.go#L38

api/v1beta1/redis_types.go
iamabhishek-dubey: Agreed

config/crd/bases/redis.redis.opstreelabs.in_redisclusters.yaml
iamabhishek-dubey: This is something the bug with operator-sdk, it removes additional printer coloums. You may have to add this manually

api/v1beta1/redis_types.go
davidcollom: I think there's confusion in this `Resources` - as its `redis.Resources` which isn't being used anywhere, but the `redis.kubernetesConfig.Resources` and `redis.redisExporter.Resources` etc. (similar for cluster and standalone.

PR #94, opened 2021-06-09
Fix issue where command output was continuously being appended

davidcollom:
Command output was being continuously appended for each execution, making logging and debugging extremely difficult to understand which portion related to what output.

This now ensures that the variables are clean on each execution.


iamabhishek-dubey (approved):
LGTM

PR #195, opened 2022-01-05
Added finalizer

MathieuCesbron:
Signed-off-by: Mathieu CESBRON <mathieu.cesbron@ecam-rennes.com>


MathieuCesbron:
With this feature, it would delete Services on deletion of CR.

PR #211, opened 2022-01-19
Simplify the calculation for Replica Size

davidcollom:
This was raised in https://github.com/OT-CONTAINER-KIT/redis-operator/pull/205#discussion_r787554381

With this change its now easier and easier to understand what size Replica should be deployed.


MathieuCesbron:
I was hesitant to add methods in [rediscluster_types.go](../blob/master/api/v1beta1/rediscluster_types.go). I think of this file as the one with all the structs that define the API, but you made me change my mind.

If we start to add helper methods in the API files, there would probably be other useful methods that could be added.

PR #214, opened 2022-01-22
Remove dead code

MathieuCesbron:
Signed-off-by: Mathieu Cesbron <mathieu.cesbron@ecam-rennes.com>


iamabhishek-dubey:
Indeed this is dead function that was created initially, thanks for removing this

PR #215, opened 2022-01-22
Simplify GetStateFulSet function

MathieuCesbron:
Hello,

Replacing `err` to `nil` when `err` is `nil` is more elegant to me.

Have a good day,


iamabhishek-dubey:
Yes as a coding practice, it should be like that because other errors are being returned in above return functions

PR #217, opened 2022-01-22
Uniformize redis reconcile

MathieuCesbron:
Hello,

We had 
 
```
if err := ...; err != nil {
}
```
and
```
err = ... 
if err != nil {
...
}
```

in redis reconcile. I prefer the first way of doing it.

Have a good day,


iamabhishek-dubey:
Looks good, can you please fix conflicts @MathieuCesbron

PR #218, opened 2022-01-22
Refacto service creation

MathieuCesbron:
Hello,

Refacto of service creation. I think it is simpler if we merge functions that create service and headlessService, tell me what you think about it.

Have a good day,


iamabhishek-dubey:
Looks good to me, actually, I did a similar kind of change in another operator. I wanted to replicate it here but could not get the time. Thanks for doing this.

PR #257, opened 2022-03-23
Make TLS finally working for RedisCluster

hufon:
TLS is now working (tested) with RedisCluster.
I've not tested for Standalone mode for now


iamabhishek-dubey (approved):
LGTM

PR #389, opened 2022-11-29
[Development][Add] Added feature for additional volume mounts

### LINKED ISSUE #321: Adding mounts to redis containers
Hey,

I would like to add modules to the cluster, I can add the `loadmodule` directive as an `additionalRedisConfig` but it needs to point to the so file on the container and I could not find a way to add any additional ConfigMap mounts to the containers.

Any direction or suggestion will be much appreciated.
Thanks!

shubham-cmyk:
Description

Adding Volume mount to the pods.

Fixes #321


shubham-cmyk:
@iamabhishek-dubey  I have opened the New PR. 
Added the Examples with snake case.

iamabhishek-dubey (approved):

PR #451, opened 2023-03-01
build: dockerfile multiplatform & caching improvements

milas:
**Description**
The most useful change here is explicitly specifying the platform for the base Go image. This enables cross-compilation using the Go toolchain.

Full list of changes:
* Use the native host Go image for cross-compiling
* Use a cache mount for Go module dependencies
* Use a cache mount for Go build cache
* ~~Use `--link` in runtime image~~
   * This prevents the need to pull `gcr.io/distroless/static` during image build if the result is being pushed (e.g. in CI)
   * I reverted this, looks like Azure Pipelines does not support this option 😢 

**Type of change**
* Build scripts improvement (sorry the other categories were really N/A)

**Checklist**
- [x] Documentation updated (N/A)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM golang:1.17 as builder
+FROM --platform=$BUILDPLATFORM golang:1.17 as builder
 ARG BUILDOS
 ARG BUILDPLATFORM
 ARG BUILDARCH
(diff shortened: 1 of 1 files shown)


iamabhishek-dubey:
Can you check why CI is failing?

milas:
I believe the CI failure is because Azure Pipelines is using the "classic" Docker builder by default; this needs features from the modern builder (BuildKit). 

I just opened #452 which is about multi-platform which also needs BuildKit enabled. There's a lot more detail there, but here's an FAQ from Azure: https://learn.microsoft.com/en-us/azure/devops/pipelines/ecosystems/containers/build-image?view=azure-devops#how-do-i-set-the-buildkit-variable-for-my-docker-builds

iamabhishek-dubey:
@milas please check CI pipeline is failing

milas:
See my comment above and #452. Closing this for now as I can't make it pass without the necessary changes to the pipeline. Feel free to ping me if you enable BuildKit on the pipeline and I will pick this up again.

PR #458, opened 2023-03-10
[Feature]: Add InitContainer Field

drivebyer:
**Description**

Add initcontainer field in cr.

Init container can do task before redis startup, such as recover task. In init container, we can share volume with redis container and prepare data volume(backup set from remote S3) for it.

Fixes #ISSUE

**Type of change**

**Checklist**

- [ ] Documentation updated


drivebyer:
PTAL @shubham-cmyk  @iamabhishek-dubey

shubham-cmyk:
It's a good idea to have an init Container for the restoration; even I was thinking the same. But we don't need here a complete container in the CR but a minimum required field. 


Here is some reference to the sidecar that we can launch
image

shubham-cmyk:
Restore and backup bash script is on the to-do list and I would be added in the next release.

drivebyer:
> But we don't need here a complete container in the CR but a minimum required field.

Fixed according to suggestion.

shubham-cmyk (commented):
Update the initContainer

shubham-cmyk (commented):
update : name init Container

PR #473, opened 2023-03-28
Update configmap example to use correct file key

artificial-aidan:
**Description**

The entrypoint script for the redis image uses `redis-external.conf` as it's external config file. But the example sets the file name to `redis-additional.conf`

**Type of change**

Documentation

### DIFF
--- a/example/additional_config/configmap.yaml
+++ b/example/additional_config/configmap.yaml
@@ -4,7 +4,7 @@ kind: ConfigMap
 metadata:
   name: redis-external-config
 data:
-  redis-additional.conf: |
+  redis-external.conf: |
     tcp-keepalive 400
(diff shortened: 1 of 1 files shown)


shubham-cmyk:
https://github.com/OT-CONTAINER-KIT/redis/blob/master/entrypoint.sh#L7

I don't think we have used  `redis-external. conf`

artificial-aidan:
I see now. It was changed in a recent version.

https://github.com/OT-CONTAINER-KIT/redis/commit/609751ff1029bbfd9de6c78df36d8195efcff4e8

This PR can be closed. It may make sense to make that file name configurable via environment variable though. Right now its not compatible with different redis versions with that entry point change.

PR #482, opened 2023-04-06
Add : CRD for the Redis-Sentinel ( Exporter ) and Update : Kustomize

sadath-12:
Added redisreplications and redissentinels path so it will be able to generate crds for them in make install command


shubham-cmyk:
@sadath-12 
Don't update make file

sadath-12:
got it

sadath-12:
check now @shubham-cmyk

PR #493, opened 2023-04-14
Bugfix on the TerminationGracePeriodSeconds change

Sasidharan3094:
**Description**

As part of the TerminationGracePeriodSeconds change, there is a small bug where if we don't specify this in manifests, the operator is failing with this error. So, fixed that and tested both with and without TerminationGracePeriodSeconds in manifests and it is working as expected. If we don't specify TerminationGracePeriodSeconds in manifests, it takes default value of 30 seconds and if specified, it takes the passed value.

Error faced when submitting a CR without TerminationGracePeriodSeconds:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1fcd63a]

Also, it would be great if we can have test cases for every code change :) 

**Checklist**

- [ ] Documentation updated


shubham-cmyk:
@Sasidharan3094  Yes there was a bug, it need to be fixed right now I have revert your previous PR. I would merge it after I have perform some check

Sasidharan3094:
> @Sasidharan3094 Yes there was a bug, it need to be fixed right now I have revert your previous PR. I would merge it after I have perform some check

Thanks @shubham-cmyk Was about to send the bugFix and will wait for your review.

Sasidharan3094:
@shubham-cmyk can you please review this

Sasidharan3094:
@shubham-cmyk The bug related is fixed, please check this pr.

shubham-cmyk:
I would do it today

Issue #507, opened 2023-05-10
[BREAKING CHANGE WITH 0.14.0] redis PODs are unable to start after updating redis-operator


chobostar:
This commit breaks compatibility with existing PODs and PVCs:
https://github.com/OT-CONTAINER-KIT/redis-operator/pull/445/files

Thus, after updating from 0.12.0 to 0.14.0, an error occurs:
>create Pod redis-prod-follower-1 in StatefulSet redis-prod-follower failed error: Pod "redis-prod-follower-1" is invalid: spec.containers[0].volumeMounts[0].name: Not found: "data"."

**How to reproduce**
1. Create persistence redis using operator version 0.12.0
2. Update operator to 0.14.0
3. Observe an error above

shubham-cmyk:
This might not happen because the CR might not be updated can you verify that.

chobostar:
> This might happen because the CR might not be updated can you verify that.

Thanks for feedback! Let me clarify how PVC works:

PVC old-name -> PV old-name = containts some data
PVC new-name -> PV new-name == empty

There is **no such thing as rename** PVC.

If POD's volume switches from "old-name" to "new-name" it will loss data.

@shubham-cmyk could you revert https://github.com/OT-CONTAINER-KIT/redis-operator/pull/445 ? It's breaking change because makes incompatible old setups with new operator.

shubham-cmyk:
Thanks, @chobostar. I would revert this PR.

chobostar:
Thanks, @shubham-cmyk 

jfyi, according to the docs:
https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names
>Most resource types require a name that can be used as a DNS subdomain name as defined in [RFC 1123](https://tools.ietf.org/html/rfc1123). This means the name must:
>contain no more than **253 characters**

K8s objects in most cases have no problem if long names.

ryan-dyer-sp:
Any chance we can get a release for this?  Seems a bit much to have this breaking bug in place for 2mo now when a fix was put in place.

PR #445, opened 2023-02-22
Optimize pvcTemplate Name for Storage #438

muicoder:
**Description**

Fixes #438 

**Type of change**

* Bug fix (non-breaking change which fixes an issue)
* New feature (non-breaking change which adds functionality)
* Breaking change (fix or feature that would cause existing functionality to not work as expected)

**Checklist**

- [x] Testing has been performed
- [x] No functionality is broken
- [ ] Documentation updated

Issue #438, opened 2023-02-18
PVC name repeated too long


muicoder:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**What version of redis-operator are you using?**

redis-operator version: v0.14.0

**Additional context**

cr.Name: `test-cluster`
```yaml
pvcName:
  Origin: test-cluster-follower-test-cluster-follower-0
  Improved: data-test-cluster-follower-0
```

drivebyer:
We could add an optional parameter to allow users to customize the PVC name prefix.

PR #510, opened 2023-05-15
Revert "Optimize pvcTemplate Name for Storage #438 (#445)"

chobostar:
This reverts commit ff6980f6bd8c1191778cc065b1d18f11f58383a7.

**Description**

Fixes [#ISSUE](https://github.com/OT-CONTAINER-KIT/redis-operator/issues/507)

**Type of change**

This revert fixes:
* breaking change of existing StatefulSet volume names
* prevent data loss of existing persistent redis deployments

Moreover there is no any "Optimize" required around PVC names:
https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names
>contain no more than **253 characters**

**Checklist**

- [x] Testing has been performed - NA
- [ ] No functionality is broken - this revert will broke deployments from 0.14.0 version
- [x] Documentation updated - NA

### DIFF
--- a/k8sutils/statefulset.go
+++ b/k8sutils/statefulset.go
@@ -298,7 +298,7 @@ func getExternalConfig(configMapName string) []corev1.Volume {
 func createPVCTemplate(stsMeta metav1.ObjectMeta, storageSpec corev1.PersistentVolumeClaim) corev1.PersistentVolumeClaim {
 	pvcTemplate := storageSpec
 	pvcTemplate.CreationTimestamp = metav1.Time{}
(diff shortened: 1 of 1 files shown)


chobostar:
Since it breaks 0.14.0 redis deployment, I could implement an optional flag for operator runtime - e.g. `--use-legacy-pvc-naming`

if `--use-legacy-pvc-naming=true`, then it name PVC as before 0.14.0. It could be some kind of compromise between redis-operator users between <0.14.0 and >=0.14.0  groups.

PR #520, opened 2023-06-13
add watchnamespace function

progmatv:
This PR add watchnamespace function. To activate it add to manager.yaml in section containers:

```
         env:
          - name: WATCH_NAMESPACE
            valueFrom:
              fieldRef:
                fieldPath: metadata.namespace
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -64,8 +64,6 @@ func main() {
 	flag.Parse()
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
-	
-	watchNamespace := getWatchNamespace()
 
 	options := ctrl.Options{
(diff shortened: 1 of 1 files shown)


shubham-cmyk:
I don't think there is a need for this PR. 
Since it is already included in the code. 

image

shubham-cmyk:
This option is already included in the code. 
https://github.com/OT-CONTAINER-KIT/redis-operator/blob/master/main.go#L77

progmatv:
OLM requires `watchNamespace` as a variable name

shubham-cmyk:
can we change the name of the env-variable here ?

progmatv:
yes this can work

PR #529, opened 2023-06-28
Persistent storage is enabled only when PVC's AccessMode is specified

muicoder:
**Description**

**Type of change**

**Checklist**

- [ ] Documentation updated


shubham-cmyk:
I think it would be better only to allow the user to create the volume with `ReadWriteOnce` or `ReadWriteOncePod`. Rather than checking the presence of the access mode.

PR #536, opened 2023-07-01
Feat: sentinel cluster size must be odd number

drivebyer:
**Description**

There is not `Not` validation in kubebuilder. We can use `Enum` to limit sentinel size to some odd number. https://book.kubebuilder.io/reference/markers/crd-validation.html

**Type of change**

**Checklist**

- [ ] Testing has been performed
- [ ] No functionality is broken
- [ ] Documentation updated

### DIFF
--- a/api/v1beta1/redissentinel_types.go
+++ b/api/v1beta1/redissentinel_types.go
@@ -7,7 +7,6 @@ import (
 
 type RedisSentinelSpec struct {
 	// +kubebuilder:validation:Minimum=1
-	// +kubebuilder:validation:Not=2
 	// +kubebuilder:default=3
 	Size                *int32                     `json:"clusterSize"`
 	KubernetesConfig    KubernetesConfig           `json:"kubernetesConfig"`


drivebyer:
PTAL @shubham-cmyk

shubham-cmyk:
I won't like to enforce an odd number of sentinels on users however having exactly 2 sentinels ( which is already taken care of )  could have conflict while making decisions but having a higher number of sentinels like 4,6,8 would be fine.

shubham-cmyk:
(For a cluster of n nodes, the quorum is n/2 + 1.)

Reference : https://martinfowler.com/articles/quorum.html

drivebyer:
hi @shubham-cmyk, according to https://redis.io/docs/management/sentinel/, the docs says:

> The quorum is the number of Sentinels that need to agree about the fact the master is not reachable, in order to really mark the master as failing, and eventually start a failover procedure if possible.
> However the quorum is only used to detect the failure. In order to actually perform a failover, one of the Sentinels need to be elected leader for the failover and be authorized to proceed. This only happens with the vote of the majority of the Sentinel processes.

So I think, the `quorum` and `majority of the Sentinel processes ` is two things for sentinel cluster.

If clustersize is 5, the majority is 3, we can tolerate 2 sentinel failed. If If clustersize is 6, the majority is 2, we can tolerate 2 sentinel failed too. That's less efficiency.

shubham-cmyk:
I think we could remove` // +kubebuilder:validation:Not=2` from the code and let the user decide the field clusterSize and quorum on it's own while providing the default value of the clusterSize=3 and quorum=2. 

@drivebyer

drivebyer:
So, we can use odd number to make sentinel cluster usage more efficient way.

drivebyer:
> I think we could remove` // +kubebuilder:validation:Not=2` from the code and let the user decide the field clusterSize and quorum on it's own while providing the default value of the clusterSize=3 and quorum=2.

Yes, of course we can.

shubham-cmyk:
Yes! We could use odd number but after 5 sentinel nodes there are very less chance of concurrent failure. so I there  would be hardly any improvement  the main issue is with exactly 2 which posses most risk of failure.

shubham-cmyk:
But I am writing validation webhook for the same where we could take care of this so right now. 
You can just remove _// +kubebuilder:validation:Not=2_.

drivebyer:
> But I am writing a validation webhook for the same where we could take care of this so right now.
> You can just remove `// +kubebuilder:validation:Not=2 `

Yes of course

PR #540, opened 2023-07-04
[Refactor] - Remove redundant finalizers from operator

muicoder:
Signed-off-by: muicoder <muicoder@gmail.com>

 because of CRs ownerReferences have already been associated

**Description**

The finalizers e.g.  `redisFinalizer` 
is added to Redis ( Custom Resource ), not the sts or the services, we already have added `ownerReferences` to the sts, service and other component which make sure deletion in every case so having function 

```
if err := finalizeRedisServices(cr); err != nil {
				return err
			}

```
or 
```

			if err := finalizeRedisStatefulSet(cr); err != nil {
				return err
			}

```
are not useful anymore because the basic operation they perform is to delete the object i.e. services and sts. 

For now we are removing the finalizers from the code as unused operations were being performed. 

**Type of change**
Refactor - This won't break the functionality of the code. 

**Checklist**
- [ ] Documentation updated

### DIFF
--- a/k8sutils/finalizer.go
+++ b/k8sutils/finalizer.go
@@ -31,13 +31,7 @@ func HandleRedisFinalizer(cr *redisv1beta1.Redis, cl client.Client) error {
 	logger := finalizerLogger(cr.Namespace, RedisFinalizer)
 	if cr.GetDeletionTimestamp() != nil {
 		if controllerutil.ContainsFinalizer(cr, RedisFinalizer) {
-			if err := finalizeRedisServices(cr); err != nil {
-				return err
-			}
 			if err := finalizeRedisPVC(cr); err != nil {
-				return err
-			}
-			if err := finalizeRedisStatefulSet(cr); err != nil {
 				return err
 			}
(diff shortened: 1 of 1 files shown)


shubham-cmyk:
@muicoder 

Why remove the finalizers

muicoder:
> @muicoder
> 
> Why remove the finalizers

CR sub-resource(StatefulSet/Service) ownerReferences have already been associated, writing here is redundant.

shubham-cmyk:
But I think adding finalizers would be a nice practice it would prevent accident deletion of the resources. 
What do you think about that ?

muicoder:
It's also a great guarantee, and the name of the resource here needs to be updated.

shubham-cmyk:
I didn't get it ?

muicoder:
The names of CR sub-resource(StatefulSet/Service) also do not match, and they are actually removed because of the ownerReferences rather than the finalizer.

shubham-cmyk:
I got your point the finalizers are not added to the sts and service.
Nice catch @muicoder 

Thanks for that

PR #549, opened 2023-07-09
Fix redis cluster pdb always come from leader params

drivebyer:


drivebyer:
PTAL @shubham-cmyk

drivebyer:
PTAL @shubham-cmyk

Issue #551, opened 2023-07-10
Hardcoded 1Mi size of node-conf PVC in RedisCluster are not allowed


genofire:
In your managed kubernetesclusters, the statefulset with there node-conf failed, now.

Minimum size in your clusters are "1Gi", please increase or make "1Mi" configurable.

Bug is there since:
#519 

please make everything configurable on this volume (also like storageclass).

shubham-cmyk:
@genofire If we make a request of 1Mi i think if it's not available it would still create a PV of 1Gi as per in my cluster.

lapete:
That depends on the cluster provider. Ionos refuses to create a volume which causes the whole deployment to fail.

The operator creates a PVC with 1Mi, no matter what is provided in the `RedisCluster` Custom Resource.

shubham-cmyk:
Thanks @lapete I think then I would make that size configurable

shubham-cmyk:
@genofire Support is enabled now

PR #519, opened 2023-06-12
Fix Bug : Operator Crash when persistence is false/Disabled

shubham-cmyk:
**Description**

Fixes: https://github.com/OT-CONTAINER-KIT/redis-operator/issues/470
Fixes: https://github.com/OT-CONTAINER-KIT/redis-operator/issues/409
Fixes : https://github.com/OT-CONTAINER-KIT/redis-operator/issues/421
Fixes : https://github.com/OT-CONTAINER-KIT/redis-operator/issues/450

_Note: The PV is created using the default storage class. If the default storage class is missing then this might throw an error._ 

**Type of change**

* Bug fix (non-breaking change which fixes an issue)
* Breaking change (fix or feature that would cause existing functionality to not work as expected)

**Checklist**

- [x] Testing has been performed
- [x] No functionality is broken
- [x] Documentation updated

PR #552, opened 2023-07-11
Fix Hardcoded 1Mi size of node-conf PVC in RedisCluster

shubham-cmyk:
Now You would be able to control the volume fully for the Node Conf. 

Fix : https://github.com/OT-CONTAINER-KIT/redis-operator/issues/551

It's an extension of PR : https://github.com/OT-CONTAINER-KIT/redis-operator/pull/519


michaelarnauts:
I think this change caused this regression: https://github.com/OT-CONTAINER-KIT/redis-operator/issues/560

There is no default `nodeConfVolumeClaimTemplate` causing existing deployments (without a nodeConfVolumeClaimTemplate) to fail when the operator is upgraded.

shubham-cmyk:
Use the Latest Helm chart (0.15.3) for this

PR #558, opened 2023-07-18
+ enhaned docs about additional config configmap format

david-pech-sestraemmy-cz:
**Description**

I have struggled a lot about how to add additional Redis configuration - this is not described well in the docs. Our use-case is to have Redis only in-memory and we need extra configuration to disable persistence. I had to go through the sources to understand that only certain key in configMap is expected. So I suggest updating the documentation and examples with this.

**Type of change**

**Checklist**

- [x] Documentation updated

### DIFF
--- a/config/crd/bases/redis.redis.opstreelabs.in_redis.yaml
+++ b/config/crd/bases/redis.redis.opstreelabs.in_redis.yaml
@@ -1501,7 +1501,8 @@ spec:
                     type: integer
                 type: object
               redisConfig:
-                description: RedisConfig defines the external configuration of Redis
+                description: RedisConfig defines the external configuration of
+                  Redis. ConfigMap with key 'redis-additional.conf' is expected.
                 properties:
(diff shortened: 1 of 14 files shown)


shubham-cmyk:
@david-pech-sestraemmy-cz 

Did you checked out https://github.com/OT-CONTAINER-KIT/redis-operator/blob/master/example/additional_config/configmap.yaml

david-pech-sestraemmy-cz:
@shubham-cmyk  What do you think?

shubham-cmyk:
I think adding this section as a comment in every manifest doesn't seem to be a good idea. 
Rather I would think to add this section in the docs as you mentioned you were not able to find that.

shubham-cmyk:
You can modify the readme.md you can add a section 
`Redis Configuration` and mention that the "Additional configuration by user can be set be passing a configmap and reference the configmap in the redis resource as per example"

Issue #555, opened 2023-07-14
Pod restarted when update annotation in custom resource


drivebyer:
**What version of redis operator are you using?**

redis-operator version: v0.14.0

**Does this issue reproduce with the latest release?**
Yes.

**What operating system and processor architecture are you using (`kubectl version`)?**

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version

</pre></details>

**What did you do?**
In my use case, I will add an annotation to Redis related custom resource, this annotation will store some description along with custom resourse. When I update the description, the redis pod restart becouse of the pod template changed.

**What did you expect to see?**
The redis pod will not restart.

**What did you see instead?**
The redis pod restarted.

drivebyer:
Maybe we should give options to add annotation filters like:
```
func filterAnnotations(anots map[string]string) map[string]string {
	// Filter out some problematic annotations we don't want in the template.
	delete(anots, "kubectl.kubernetes.io/last-applied-configuration")
	delete(anots, "banzaicloud.com/last-applied")
	return anots
}
```

drivebyer:
@shubham-cmyk  @iamabhishek-dubey

shubham-cmyk:
Okay so you don't want these specific annotations to be copied over the pods and stateful set. 
The above idea is good. I would open a PR specific to that.

shubham-cmyk:
I see that it is already implemented. 
Code : https://github.com/OT-CONTAINER-KIT/redis-operator/blob/master/k8sutils/labels.go#L93

refer : https://github.com/OT-CONTAINER-KIT/redis-operator/pull/193

drivebyer:
I am still facing this issue because in my use case, I am adding an additional annotation to a custom resource which is used to store a description and will be updated by the user. I want to ensure that this annotation is not transferred to the Pod template.

@shubham-cmyk

shubham-cmyk:
Got it you are using some custom annotation.

broken-ufa:
Hi! I also need facing this issue and waiting for fix. Thank you

PR #565, opened 2023-07-25
Ignore annotatoins on sts, pods and svc

shubham-cmyk:
Fixes : https://github.com/OT-CONTAINER-KIT/redis-operator/issues/555


drivebyer:
This pull request doesn't seem to update the definition of the CRD.

PR #571, opened 2023-08-02
Migrate the Pipeline from Azure to Github actions

shubham-cmyk:
**Description**

Fixes https://github.com/OT-CONTAINER-KIT/redis-operator/issues/564, https://github.com/OT-CONTAINER-KIT/redis-operator/issues/504, https://github.com/OT-CONTAINER-KIT/redis-operator/issues/452

**Type of change**

1. This PR would address the release as mentioned by https://github.com/OT-CONTAINER-KIT/redis-operator/issues/527
Release of the appVersion would only happen if not present the `latest` tag image would be updated on every merge. 

2. This PR also fixed the Multi arch image redis-operator support 

Not the image of the redis-operator would be pushed like : 

latest: Multi arch image that supports both platforms 
v0.15.0: Multi arch image that supports both platforms 
v0.15.0-amd64 : Only for amd64
v0.15.0-arm64: Only for arm64

**Checklist**

- [ ] Testing has been performed
- [ ] No functionality is broken
- [ ] Documentation updated

PR #572, opened 2023-08-02
Close redis client to avoid resource leak

drivebyer:


shubham-cmyk:
This would be merge after the https://github.com/OT-CONTAINER-KIT/redis-operator/pull/571

PR #579, opened 2023-08-10
Add - GithubAction For Incoming e2e Test Case

drivebyer:
Adding GithubAction that would test the redis-operator image on the kind cluster


shubham-cmyk:
@drivebyer It is not failing due to that. We have to make a proper KIND related setup. I have tested it on a live cluster it was working there. KIND setup is what I need to achieve. 

I would push some changes regards that

drivebyer:
> 

OK

shubham-cmyk:
@drivebyer For this change we should not use the helm chart for deploying the operator instead we should build the operator right from scratch( Dockerfile )  from here and install it. 

We could look how the make install would work here

drivebyer:
> @drivebyer For this change we should not use the helm chart for deploying the operator instead we should build the operator right from scratch( Dockerfile ) from here and install it.
> 
> We could look how the make install would work here

Yes, I am testing now. I could build image on runner, then use `kind load` load the fresh operator image to kind nodes.

PR #584, opened 2023-08-16
[Feature] : Add support for multiple versions

### LINKED ISSUE #566: Regression when using pod security context
This [commit](https://github.com/OT-CONTAINER-KIT/redis-operator/commit/2142b250aa4fa22458504cc6bf2e62ece4f2b251) renamed the field "securityContext" to "podSecurityContext" breaking backward compatibility.

This means that charts built on top of this CRD with 0.14.0 cannot be used with 0.15.0. The following should be done:

1. Revert the name change to the previous name securityContext for the PodSecurityContext and another name for the "real" SecurityContext type (IE. containerSecurityContext)

If you want to change names, you should produce a new version of the CRD object then it doesn't break compatibility along with a webhook to handle conversion

It might end up duplicating effort to produce new CRD version, but that would the less impacting path to support upgrade using more recent versions of the operator while still allowing bigger changes toward a 1.0 version

My day job has many teams using this operator and upgrading 0.15 will require quite the effort since our deployment are managed by argoCD, meaning that we can't just update the operator because if we delete resources after tagging them with reconcile-skip, they are going to be recreated from the code repository code preventing such configuration to be created as they won't match the new CRD.

We can potentially contribute with a PR depending on which approach you would prefer for the fix.

shubham-cmyk:
**Description**

This PR would add support so that we could have multiple versions of the crd's right now adding `v1beta2`. 

`v1beta2` would cover the scheme followed by the v0.15.0 
`v1beta1` would cover the scheme followed by the v0.14.0 
v1beta2 would be stored 
and both versions would be served. 

This PR would also help us to migrate the resources and would be needed as v0.14.0 and v0.15.0 are not compatible. 
Scheme change could be handled. 

Fixes #566 

**Type of change**

**Checklist**

- [ ] Testing has been performed
- [ ] No functionality is broken
- [ ] Documentation updated

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -11,6 +11,10 @@ rules:
   - redisclusters
   - redis
   - rediscluster
+  - redisreplication
+  - redisreplications
+  - redissentinel
+  - redissentinels
   verbs:
   - create
   - delete


shubham-cmyk:
@iamabhishek-dubey 
can you have a look

darkrift:
How will that support the PVC split ?

shubham-cmyk:
> How will that support the PVC split ?

The new PVC would created by the operator and also you need to use new redis-image to support that split

shubham-cmyk:
Follow up : https://github.com/OT-CONTAINER-KIT/redis-operator/pull/592

Issue #430, opened 2023-02-16
Log pollution 


shubham-cmyk:
**Is your feature request related to a problem? Please describe.**
Log are over polluted 

**Describe the solution you'd like**
I would like to make the logs of the redis-operator a little better to reduce unnecessary logs

**Describe alternatives you've considered**

**What version of redis-operator are you using?**

redis-operator version: v0.14.0

**Additional context**

shubham-cmyk:
@drivebyer 
You can look at this

PR #585, opened 2023-08-19
[Enhancement]- Fix log pollution

drivebyer:
**Description**

Fixes https://github.com/OT-CONTAINER-KIT/redis-operator/issues/430

Add log verbosity param && level down some logger.

**Type of change**

**Checklist**

- [x] Documentation updated

### DIFF
--- a/main.go
+++ b/main.go
@@ -20,6 +20,8 @@ import (
 	"flag"
 	"os"
 
+	"go.uber.org/zap/zapcore"
+
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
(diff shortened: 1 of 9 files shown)


shubham-cmyk:
@drivebyer 
If we change the level for the logger from here https://github.com/OT-CONTAINER-KIT/redis-operator/blob/master/k8sutils/redis.go#L431

It would be much simpler to handle

drivebyer:
> If we change the level for the logger from here https://github.com/OT-CONTAINER-KIT/redis-operator/blob/master/k8sutils/redis.go#L431

yes, I have already change the logger level from ZERO to ONE, along with many other loggers.

main.go
shubham-cmyk, line 1: In this file you should also add a way to control the verbosity via env var `LOG_LEVEL` could be name and args could be `--log-level`

shubham-cmyk (commented):
Also @drivebyer Changing log level won't just solve the problem you have to move many of the log that we were marking on the info to the debug

main.go
drivebyer: Actually, i find the `zap-log-level` already added by `opts.BindFlags(flag.CommandLine)`.

shubham-cmyk: Okay why you removed that ? 
You should add that

shubham-cmyk (approved):
@drivebyer Great Work! Thanks a lot for PR

PR #617, opened 2023-09-10
Support ENABLE_WEBHOOKS env which cloud disable webhook server locally

drivebyer:
**Description**

When run operator locally, we may not use webhook server usaually, then we can provide option to disable webhook server, just like: https://github.com/kubernetes-sigs/controller-runtime/issues/491#issuecomment-617028704

**Type of change**

**Checklist**

**Additional Context**

### DIFF
--- a/main.go
+++ b/main.go
@@ -122,21 +122,24 @@ func main() {
 		setupLog.Error(err, "unable to create controller", "controller", "RedisSentinel")
 		os.Exit(1)
 	}
-	if err = (&redisv1beta2.Redis{}).SetupWebhookWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create webhook", "webhook", "Redis")
-		os.Exit(1)
-	}
-	if err = (&redisv1beta2.RedisCluster{}).SetupWebhookWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create webhook", "webhook", "RedisCluster")
-		os.Exit(1)
-	}
-	if err = (&redisv1beta2.RedisReplication{}).SetupWebhookWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create webhook", "webhook", "RedisReplication")
-		os.Exit(1)
-	}
-	if err = (&redisv1beta2.RedisSentinel{}).SetupWebhookWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create webhook", "webhook", "RedisSentinel")
-		os.Exit(1)
+
+	if os.Getenv("ENABLE_WEBHOOKS") != "false" {
+		if err = (&redisv1beta2.Redis{}).SetupWebhookWithManager(mgr); err != nil {
+			setupLog.Error(err, "unable to create webhook", "webhook", "Redis")
+			os.Exit(1)
+		}
+		if err = (&redisv1beta2.RedisCluster{}).SetupWebhookWithManager(mgr); err != nil {
+			setupLog.Error(err, "unable to create webhook", "webhook", "RedisCluster")
+			os.Exit(1)
+		}
+		if err = (&redisv1beta2.RedisReplication{}).SetupWebhookWithManager(mgr); err != nil {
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -122,21 +122,24 @@ func main() {
...
-	}
-	if err = (&redisv1beta2.RedisReplication{}).SetupWebhookWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create webhook", "webhook", "RedisReplication")
-		os.Exit(1)
-	}
-	if err = (&redisv1beta2.RedisSentinel{}).SetupWebhookWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create webhook", "webhook", "RedisSentinel")
-		os.Exit(1)
+
+	if os.Getenv("ENABLE_WEBHOOKS") != "false" {
```
shubham-cmyk, line 126: What would happen in 4 cases 
1. I set the ENABLE_WEBHOOKS=true
2. I set the ENABLE_WEBHOOKS=false
3. I didn't set this env variable 
4. I set ENABLE_WEBHOOKS=abc which could be some arbitrary value.

This would fail to execute in the 4th pt. make sure you fix it and check the result. 

The default behavior of  `didn't set this env variable` should enable the webhook. 

PS> Make sure you add this change to the helm chart also.

shubham-cmyk: @drivebyer 
Could we take an arg from user for this also 
like 
--webhooks=false or --webhook=true

drivebyer: > * I set the ENABLE_WEBHOOKS=true
> * I set the ENABLE_WEBHOOKS=false
> * I didn't set this env variable
> * I set ENABLE_WEBHOOKS=abc which could be some arbitrary value.

All following commands execute out of k8s cluster(as locally)

1. execute `ENABLE_WEBHOOKS=true make run`, got error log `{"level":"error","ts":1694398587.147439,"logger":"setup","msg":"problem running manager","error":"open /var/folders/w_/5wv_1bls2rq8dt_vclll23j00000gn/T/k8s-webhook-server/serving-certs/tls.crt: no such file or directory","stacktrace":"main.main\n\t/Users/wuyang/workspace/github.com/drivebyer/redis-operator-1/main.go:157\nruntime.main\n\t/Users/wuyang/sdk/go1.17.13/src/runtime/proc.go:255"}`
2. execute `ENABLE_WEBHOOKS=false make run`, the operator start normally
3. the same as 1
4. the same as 1

drivebyer: > PS> Make sure you add this change to the helm chart also.

IMO, this environment flag just for running operator on locally for debugging, When deploy operator to k8s cluster, we always need webhook server enabled. So I think we don't need add this environment flag to helm chart.

drivebyer:
BTW, I found these:
1. https://github.com/OpenFunction/OpenFunction/blob/92fe37b7b6f4b4e9d75f2e2cca050958787a5f99/main.go#L185-L202
2. https://github.com/apache/skywalking-swck/blob/fd6f354bf8af88d0196d02ff5ca2b8be41a847b1/operator/main.go#L171
3. https://github.com/open-telemetry/opentelemetry-operator/blob/main/main.go#L246

main.go
drivebyer: > Could we take an arg from user for this also
> like
> --webhooks=false or --webhooks=true

Yes We could

shubham-cmyk (approved):
Lgtm

main.go
shubham-cmyk: > 1. {"level":"error","ts":1694398587.147439,"logger":"setup","msg":"problem running manager","error":"open /var/folders/w_/5wv_1bls2rq8dt_vclll23j00000gn/T/k8s-webhook-server/serving-certs/tls.crt: no such file or directory","stacktrace":"main.main\n\t/Users/wuyang/workspace/github.com/drivebyer/redis-operator-1/main.go:157\nruntime.main\n\t/Users/wuyang/sdk/go1.17.13/src/runtime/proc.go:255"}

It failed to the the tls.crt and tls.key not passed here. 
I have mounted these volume in the helm chart of redis-operator

PR #661, opened 2023-10-06
[Test] : e2e Github Action Workflow 

shubham-cmyk:
This is the correct version of: https://github.com/OT-CONTAINER-KIT/redis-operator/pull/579/files

Description : 

1. Every test dir would be overwritten in this workflow and should be launched a separate action
2. Docker image would be built from ./Dockerfile 
3. Operator would be installed from the present manifest in the dir : config

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       matrix:
         testpath:
-        - ./e2e/v1beta2/setup
+        - .tests/e2e/v1beta2/setup
 
     steps:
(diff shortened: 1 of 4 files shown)


shubham-cmyk:
@drivebyer 

1. We have disabled webhook here
2. make deploy only used redis-operator:e2e Image for now. 


These need to be fixed in future

drivebyer:
> @drivebyer
> 
> 1. We have disabled webhook here
> 2. make deploy only used redis-operator:e2e Image for now.
> 
> These need to be fixed in future

OK, I will create two issue to track it.

PR #673, opened 2023-10-21
feat:  verify codgen in Makefile and Github workflow

shubham-cmyk:
verify that the proper code generation is happening or not. 
Fixing the Improper variable name in the workflow.

Now we could use 

`make codegen` - To generate crd + deepcopy 
`make verify-codegen` - To verify if we have any any missing crd ( any other manifests ) or deepcopy

### DIFF
--- a/.github/workflows/operator-ci.yaml
+++ b/.github/workflows/operator-ci.yaml
@@ -8,11 +8,11 @@ on:
       - master
 
 env:
-  GolangVersion: 1.21
-  ApplicationName: redis-operator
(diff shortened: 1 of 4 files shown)


shubham-cmyk:
@drivebyer Can you check this ?

drivebyer:
> @drivebyer Can you check this?
> 
> This seems to be important to me since the last time in the API change you didn't pushed deep copy out there, which has caused issues. Now we could make sure user has to push the required code-gen

LGTM, it's nice!

PR #682, opened 2023-10-28
[Fix] : redis cluster status after scale up

drivebyer:
1.  Fix redis cluste ready node no changed afte scale up
2. Add e2e test for scale up


shubham-cmyk:
@drivebyer 

Can we also make sure here that the scale-down is working properly here ? 

We could have another `02-scale-down.yaml`

drivebyer:
> @drivebyer
> 
> Can we also make sure here that the scale-down is working properly here ?
> 
> We could have another `02-scale-down.yaml`

I would add it, but I think it should be in different PR.

PR #707, opened 2023-11-23
feat(client): generate clientset informer lister

elrondwong:
add client #706


shubham-cmyk:
test is failing @drivebyer doesn't seems right

drivebyer:
@elrondwong  Could you fix the `govet` && `Go Test` error?

PR #726, opened 2023-12-12
test: try Ping cluster

shubham-cmyk:
Signed-off-by: Shubham Gupta <iamshubhamgupta2001@gmail.com>


shubham-cmyk:
@drivebyer 
Check this out we have to use the literal over here not the string. 
The error was with the "PONG" it should be 'PONG'

PR #746, opened 2024-01-10
docs: add namespace creation

MathieuCesbron:
**Description**

If we follow the installation guidelines, it failed because the namespace `ot-operators` does not exist. We add a line to create it in the documentation.

**Type of change**

**Checklist**

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,7 +12,7 @@
 
 ## Local Kubernetes Cluster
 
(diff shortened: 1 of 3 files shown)


drivebyer:
how about use `helm upgrade --create-namespace -n ot-operators`? then can combine to one command.

MathieuCesbron:
Thanks for the suggestion, it is indeed better this way ! I just updated the commit, it should be good.

I added the namespace creation every time we `helm upgrade` the operator.

PR #748, opened 2024-01-11
test: add envtest for redis standalone

MathieuCesbron:
**Description**

I feel like it would be a good idea to add integration tests with EnvTest on the reconcile loop [like mentioned here](https://www.infracloud.io/blogs/testing-kubernetes-operator-envtest/).

Would this feature be accepted if I make it ?

_Edit: Feature accepted_

Fixes #656

**Type of change**

* Bug fix (non-breaking change which fixes an issue)

**Checklist**

- [x] Tests have been added/modified and all tests pass.
- [x] Functionality/bugs have been confirmed to be unchanged or fixed.
- [x] I have performed a self-review of my own code.
- [x] Documentation has been updated or added where necessary.

PR #750, opened 2024-01-13
feat: generate k8s client only once

MathieuCesbron:
**Description**

I realised that we create a new k8s client every time we need it. It's quite consuming, we should create it at the start of the controller and use it for his lifetime. We then pass the client down to each functions.

I realised this because we cannot add integration tests with EnvTest if we create a new client every time we need one #748

PS: It looks like the coverage test fail because some code removed was tested so it technically lowers the coverage.

**Type of change**

**Checklist**


drivebyer (approved):
lgtm

PR #760, opened 2024-01-25
chore: remove latest version for conversion webhook

MathieuCesbron:
**Description**

The point of the conversion webhook is too update some v1beta1 fields basically and it does nothing for v1beta2. I removed the conversion webhook for v1beta2 version because there is no need to call the conversion for them.

**Type of change**

**Checklist**

### DIFF
--- a/config/crd/patches/webhook_in_redis.yaml
+++ b/config/crd/patches/webhook_in_redis.yaml
@@ -12,4 +12,4 @@ spec:
           namespace: ot-operators
           name: webhook-service
           path: /convert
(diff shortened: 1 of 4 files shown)


drivebyer:
Some people may still use v1beta1. IMO, we still need conversion in a period of time.

MathieuCesbron:
Thanks for the comment. Yes, I agree we should only trigger the conversion for v1beta1 version and not for v1beta2 because it’s gonna convert nothing.

That’s why I removed it from the selector. Am I missing something ?

drivebyer (approved):

PR #767, opened 2024-02-01
chore: enhance log every reconcile redis cluster

### LINKED ISSUE #764: "Creating redis cluster by executing cluster creation commands" log line is redundand and misleading
Source code link:

https://github.com/OT-CONTAINER-KIT/redis-operator/blob/3e6fcb9a5f44052db0767cbe278f83516d278741/controllers/rediscluster_controller.go#L194

Operator issues "Creating redis cluster by executing cluster creation commands" every reconciliation cycle (i.e. every 10 seconds).

This is extremely misleading as it creates perception that cluster is NOT created/reconciled yet and extra actions are still necessary, whereas in reality the cluster has been ready for some time.

It also does not add any value during actual cluster configuration phase as it's not directly linked to any operations

drivebyer:
**Description**

close #764

### DIFF
--- a/controllers/rediscluster_controller.go
+++ b/controllers/rediscluster_controller.go
@@ -191,8 +191,8 @@ func (r *RedisClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


shubham-cmyk (approved):
LGTM

PR #795, opened 2024-03-02
ci: remove strategy from test

shubham-cmyk:
**Description**

Consider removing strategy from the CI since chainsaw can read the nested directory. 

I am not sure we should do it or not it. We can consider this since it would less github runner. 
we can make the test run in parallel that is done by default. 

WDYT ? 
@drivebyer

### DIFF
--- a/.github/workflows/e2e-chainsaw.yml
+++ b/.github/workflows/e2e-chainsaw.yml
@@ -22,23 +22,8 @@ jobs:
       run: yamllint --strict ./tests/
 
   test:
-    name: ${{ matrix.testpath }}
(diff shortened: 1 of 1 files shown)


drivebyer:
I believe it would be beneficial if Chainsaw could support the parallel execution of test cases. This way, we might be able to run all tests on a single runner.

shubham-cmyk:
@drivebyer chainsaw actually execute the test in the parallel. 
we can control this

drivebyer:
done by https://github.com/OT-CONTAINER-KIT/redis-operator/pull/1066

PR #806, opened 2024-03-07
fix: sentinel waiting for replication pod got ready

drivebyer:
**Description**

We should wait for the replication cluster to be ready before creating the sentinel cluster.

**Type of change**

**Checklist**

**Additional Context**

### DIFF
--- a/tests/e2e-chainsaw/v1beta2/teardown/redis-sentinel/sentinel.yaml
+++ b/tests/e2e-chainsaw/v1beta2/teardown/redis-sentinel/sentinel.yaml
@@ -8,8 +8,9 @@ spec:
   podSecurityContext:
     runAsUser: 1000
     fsGroup: 1000
-  redisSentinelConfig:
-    redisReplicationName : redis-replication
+   # If we specify the redisReplicationName, we should wait for the redis cluster to be ready before deploying the sentinel
+#  redisSentinelConfig:
+#    redisReplicationName : redis-replication
   kubernetesConfig:
     image: quay.io/opstree/redis-sentinel:latest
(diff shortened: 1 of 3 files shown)


tests/e2e-chainsaw/v1beta2/teardown/redis-sentinel/sentinel.yaml
```diff
@@ -8,8 +8,9 @@ spec:
   podSecurityContext:
     runAsUser: 1000
     fsGroup: 1000
-  redisSentinelConfig:
-    redisReplicationName : redis-replication
+   # If we specify the redisReplicationName, we should wait for the redis cluster to be ready before deploying the sentinel
```
shubham-cmyk, line 11: Why don't we use the chinsaw wait operation in another step so that we can wait till the redis replication is ready @drivebyer

drivebyer: > Why don't we use the chainsaw wait operation in another step so that we can wait till the redis replication is ready @drivebyer

In my opinion, in this test case, it is sufficient to install Sentinel independently, and then test the StatefulSet and Service. We do not specify redisReplicationName, the Sentinel StatefulSet and Service will still be created as well.

drivebyer: We have already test RedisReplication with RedisSentinel in https://github.com/OT-CONTAINER-KIT/redis-operator/tree/master/tests/e2e-chainsaw/v1beta2/ha-setup :) @shubham-cmyk

PR #840, opened 2024-03-23
chore: use redis client library

shubham-cmyk:


shubham-cmyk:
@drivebyer 
Can you approve this

drivebyer:
Great! This is a more elegant way.

PR #841, opened 2024-03-23
test(client): Add redis mock client

shubham-cmyk:
Add redis mock client for testing the result from the redis


shubham-cmyk:
@drivebyer 
PTAL

drivebyer:
lgtm.

In this way, we can test more end-to-end cases with fewer end-to-end actions.

PR #850, opened 2024-03-29
ci: run e2e after CI workflow completed

drivebyer:
E2E workflow can be run after checking that the operator CI is successful


shubham-cmyk:
@drivebyer The e2e test didn't ran here

shubham-cmyk:
e2e Test Still not running

drivebyer:
@shubham-cmyk Yes, I noticed this. I can't figure out why it's not working. Make it drafting now, maybe need some time to figure it out.

PR #1023, opened 2024-07-10
feat: support setting minReadySeconds on the stateful sets

### LINKED ISSUE #1022: stateful set minReadySeconds
I'm looking to use this operator to deploy replication+sentinal and I need to avoid what I'm seeing in terms of replication collapsing due to pods restarting faster than sentinel and the operator can handle.
The HA in this model works reasonably well if you restart a single pod, which is great.
But what if you're doing a node cycle on the cluster, or just restarting the stateful set.  My tests indicate that the replication falls apart and has to be re-initialised.
So I suggest that the operator is extended to configure minReadySeconds on the stateful sets which has been standard since k8s 1.25 https://kubernetes.io/blog/2022/09/15/app-rollout-features-reach-stable/#solved-problem-statefulset-minreadyseconds
That way there is a delay between pod restarts with a replicating stateful set and there is time for sentinel etc to do their work failing over during the process.
Happy to raise a PR for this if it would be welcomed.

mattrobinsonsre:
**Description**

minReadySeconds has been standard on stateful sets since v1.25 https://kubernetes.io/blog/2022/09/15/app-rollout-features-reach-stable/#solved-problem-statefulset-minreadyseconds
Configuring it allows one to control the rate at which pods are replaced.  This is especially important for a replication+sentinel configuration where there needs to be a reasonable amount of time allowed for the failover processes to complete between pod replacements.

Fixes #1022 

feature

**Checklist**

**Additional Context**


mattrobinsonsre:
This probably needs more work,  I couldn't find a suitable test to update.  Would appreciate any pointers.

mattrobinsonsre:
Found and updated the tests and also updated the examples.  Tests pass

drivebyer:
Thanks for your contribution. @mattrobinsonsre

drivebyer (approved):

drivebyer:
This PR causes backward compatibility problems, like https://github.com/OT-CONTAINER-KIT/redis-operator/issues/1030#issuecomment-2244847108. We need to set `minReadySeconds` as an optional value in the chart.

PR #1038, opened 2024-08-02
ping instead of checking pod readiness

### LINKED ISSUE #1016: Using a readiness probe halts operator progress
It seems like we check the ready state of all of the replicas here: https://github.com/OT-CONTAINER-KIT/redis-operator/blob/08eb5eb2364aaa77f62b4b2740d17863c956deef/k8sutils/statefulset.go#L77

When you use a custom readiness probe though that checks cluster status, the operator can get into a stuck state during bootstrapping and scale in/out since the readiness probe will be failing until the operator actually takes action to create the cluster or add/remove nodes from the cluster.

Is there any way we can base the operator off of the liveness probes, or maybe only take into account readiness when we expect the cluster to be in a ready state?

com6056:
**Description**

Right now, since the operator relies on checking pod readiness, it makes using a more robust readiness probe not possible. This changes it so instead of checking the pod readiness, we just directly `PING` all of the redis nodes instead (which the default readiness probe does anyways).

Fixes #1016 

**Type of change**

**Checklist**

**Additional Context**


drivebyer:
@com6056 Could you please resolve the conflict

PR #1261, opened 2025-03-04
fix: replace hardcoded Redis port 6379 with configurable port from cr.Spec.Port

fengyinqiao:
### Changes
- Updated `RedisClusterStatusHealth` function to use `cr.Spec.Port` instead of hardcoded 6379 port.
- Ensured the command uses the specified port from the RedisCluster spec.
- Added handling for existing password secret and TLS arguments.

### Context
This change allows the Redis port to be configurable through the `cr.Spec.Port` field, improving flexibility and configurability of the Redis cluster setup.

### Testing
- Verified that the Redis cluster health check uses the specified port from the RedisCluster spec.
- Ensured that the command execution works correctly with the provided port, password, and TLS arguments.


fengyinqiao:
@drivebyer Please take a look at this fix, you wrote this code.

drivebyer (approved):
lgtm

PR #1180, opened 2024-12-20
feat: redis-cluster add podAntiAffinity(#1174 )

xiaozhuang-a:
**Description**

support leader and follower anti affinity of pods with the same order

Fixes #ISSUE

**Type of change**

* New feature (non-breaking change which adds functionality)

**Checklist**

- [ ] Tests have been added/modified and all tests pass.
- [ ] Functionality/bugs have been confirmed to be unchanged or fixed.
- [x] I have performed a self-review of my own code.
- [ ] Documentation has been updated or added where necessary.

**Additional Context**

PR #1296, opened 2025-03-24
feat: Add support for anti affinity configuration in helm charts

NotAndD:
**Description**

Following what was done in [pull/1180](https://github.com/OT-CONTAINER-KIT/redis-operator/pull/1180), there is an annotation that can be used to enforce leader-follower anti affinity using a mutating webhook. But it's not possible to use it from the provided Helm Charts.

I added the following:

- Possibility to add the annotation on the `RedisCluster` resource by using `redisCluster.enableMasterSlaveAntiAffinity`
- Possibility to disable the creation of cert-manager `Issuer` (which might be not required) in case someone wants to use an existing one for the `Certificate`
- Possibility to create it as a `ClusterIssuer`, if desired. And also to use this kind of issuer in the certificate.

**Type of change**

**Checklist**

    - There are no tests for the Helm Charts (it seems)

**Additional Context**

I left the default parameters in the values to configure the chart as before, so there should be no changes, unless parameters are personalized.

### DIFF
--- a/charts/redis-cluster/README.md
+++ b/charts/redis-cluster/README.md
@@ -70,6 +70,7 @@ helm delete <my-release> --namespace <namespace>
 | priorityClassName | string | `""` |  |
 | redisCluster.clusterSize | int | `3` |  |
 | redisCluster.clusterVersion | string | `"v7"` |  |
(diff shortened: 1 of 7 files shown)


NotAndD:
Sorry, thought the readme files were updated manually, should have noticed the templetized readmes (internal facepalm)

drivebyer (approved):
lgtm

Issue #1313, opened 2025-04-01
Why do we copy labels & annotations from the StatefulSet to the PVC?


seeker89:
**Describe your question**

While debugging an issue, I noticed that we copy the labels & (most) annotations from the `statefulset` to the `PVC`, as in here:

https://github.com/OT-CONTAINER-KIT/redis-operator/blob/5ea1410f20b31b0711b154fa1d40c0d42f610d23/pkg/k8sutils/statefulset.go#L379-L384

If the resource is managed by other actors (e.g. `ArgoCD`), incorrect IDs are copied over. 

What is the rationale for doing so?

**What version of redis-operator are you using?**

redis-operator version: `v0.19.1` and just now `v0.20.0`

**Additional context**

If the reason is to have an easy way of passing the metadata to the PVC, perhaps we should only copy over things matching a specified prefix?

PR #1322, opened 2025-04-08
fix: move VCT logic before diff calculation for stateful set

### LINKED ISSUE #1321: Infinite reconciliation loop in StatefulSet never recovering

redis-operator version: `v0.20.0`

yes

**What did you do?**

To reproduce this bug, follow these steps:

* create a `RedisReplication` CRD & wait for it to create the corresponding `StatefulSet`
* add a new annotation to the `RedisReplication` CRD (any)
* watch it go into an infinite loop unable to reconcile the `VolumeClaimTemplate` part

No infinite loop.

**What did you see instead?**

The operator hammering the `kube-apiserver` with infinite loop of updates.

seeker89:
### How to reproduce the bug

1. create a `RedisReplication` CRD & wait for it to create the corresponding `StatefulSet`
1. add a new annotation to the `RedisReplication` CRD (anything will do)
1. watch it go into an infinite loop unable to reconcile the `VolumeClaimTemplate` part

With the default log level it's not initially obvious what's happening, so it might be worth adding `--zap-devel` to the argument list.

You will notice the `banzaicloud.com/last-applied` annotation changing. `kubectl get sts --watch` and/or `k8s` audit logs all will show the continuous updates.

### Why the issue occurs

The issue is that the diffing logic happens before the VCT modification happens.

So this calculates a diff which includes e.g a new annotation, and the patch is non-empty:

https://github.com/OT-CONTAINER-KIT/redis-operator/blob/5ea1410f20b31b0711b154fa1d40c0d42f610d23/pkg/k8sutils/statefulset.go#L183-L200

However, later it is acknoledges that VCT is immutable anyway, and it's excluded from the request that goes on to be sent for update:

https://github.com/OT-CONTAINER-KIT/redis-operator/blob/5ea1410f20b31b0711b154fa1d40c0d42f610d23/pkg/k8sutils/statefulset.go#L212-L217

Because the change is excluded, the only thing that really changes is the `banzaicloud.com/last-applied` and in the next loop iteration we go back and see the same patch, and so on.

### Description

This PR fixes this issue in the least disruptive manner I could think of this morning.

To make things more confusing, the code contains a red herring:

https://github.com/OT-CONTAINER-KIT/redis-operator/blob/5ea1410f20b31b0711b154fa1d40c0d42f610d23/pkg/k8sutils/statefulset.go#L191

This sounds like just what you need, but unfortunately the implementation only resets some select fields:

https://github.com/banzaicloud/k8s-objectmatcher/blob/a7fe438786b4ddeaf76c56d430ac725cdfa3cfc7/patch/deletenull.go#L222-L245

```go
func deleteVolumeClaimTemplateFields(obj []byte) ([]byte, error) {
	resource := map[string]interface{}{}
	err := json.Unmarshal(obj, &resource)
	if err != nil {
		return []byte{}, errors.Wrap(err, "could not unmarshal byte sequence")
	}

	if spec, ok := resource["spec"]; ok {
		if spec, ok := spec.(map[string]interface{}); ok {
			if vcts, ok := spec["volumeClaimTemplates"]; ok {
				if vcts, ok := vcts.([]interface{}); ok {
					for _, vct := range vcts {
						if vct, ok := vct.(map[string]interface{}); ok {
							vct["kind"] = ""
							vct["apiVersion"] = ""
							vct["status"] = map[string]string{
								"phase": "Pending",
							}
						}
					}
				}
			}
		}
	}
```

⚠️ a quick note aside: for each of this ignoring functions, it unmarshals and marshals JSON twice, which is super wasteful.

So this PR basically moves the VCT logic up, but keeps a copy of the unchanged new state, to make sure we don't change the behaviour of `HandlePVCResizing` (which would make this PR more involved).

Fixes #1321

### Type of change

### Checklist

### DIFF
--- a/pkg/k8sutils/statefulset.go
+++ b/pkg/k8sutils/statefulset.go
@@ -185,6 +185,18 @@ func patchStatefulSet(ctx context.Context, storedStateful, newStateful *appsv1.S
 	// Sync system-managed fields to ensure atomic update.
 	syncManagedFields(storedStateful, newStateful)
 
+	// Save the new VolumeClaimTemplates for later use in HandlePVCResizing
+	newStatefulCopy := newStateful.DeepCopy()
+
+	// Since VolumeClaimTemplate fields are immutable, revert to the stored configuration.
+	if hasVolumeClaimTemplates(newStateful, storedStateful) {
+		if newStateful.Annotations == nil {
+			newStateful.Annotations = make(map[string]string)
+		}
+		newStateful.Annotations["storageCapacity"] = storedStateful.Annotations["storageCapacity"]
+		newStateful.Spec.VolumeClaimTemplates = storedStateful.Spec.VolumeClaimTemplates
+	}
+
 	// Calculate the patch between the stored and new objects, ignoring immutable or unnecessary fields.
 	patchResult, err := patch.DefaultPatchMaker.Calculate(storedStateful, newStateful,
 		patch.IgnoreStatusFields(),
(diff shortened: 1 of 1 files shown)


seeker89:
I've updated the title so that your CI is happy.

Also, this previous issue is related - I'm still not sure why the operator is copying all the annotations and labels, and I suspect this behaviour might cause more trouble in the future https://github.com/OT-CONTAINER-KIT/redis-operator/issues/1313

seeker89:
Hello - have you had a chance to take a look?

dimpavloff:
Thanks for the fix @seeker89 !
@drivebyer , would you be up for cutting a release with this change please?

drivebyer:
> Thanks for the fix @seeker89 ! @drivebyer , would you be up for cutting a release with this change please?

Sure

dimpavloff:
Thanks!

PR #1330, opened 2025-04-28
fix: handle panic when retrieving StatefulSet in GetRedisNodesByRole

shubham-cmyk:
Signed-off-by: Shubham Gupta <iamshubhamgupta2001@gmail.com>

**Description**

Test were panic due to continuation on the error.
We need to return if we have found the error

```
Will run 1 of 1 specs
•panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x188ba15]

goroutine 251 [running]:
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile.func1()
	/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.17.2/pkg/internal/controller/controller.go:116 +0x1da
```

**Type of change**

**Checklist**

**Additional Context**

Tried Running Test More that 1 one to make sure to catch the flaky condition

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -180,7 +180,7 @@ verify-codegen: codegen
 # ===========================
 
 .PHONY: tests
-tests: integration-test-setup unit-tests
+tests: integration-test-setup unit-tests integration-tests
 
 .PHONY: unit-tests
 unit-tests:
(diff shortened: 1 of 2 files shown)


shubham-cmyk:
@drivebyer 

I have looked at the flake maybe sometime the sts is not created and the continuation happen which cause the panic since we try to get the name of the sts from the ptr. 

This could be a fix.

drivebyer (approved):
lgtm

PR #1342, opened 2025-05-07
fix: VCT resize detection logic; add support for scaling out with new VCT size

### LINKED ISSUE #1321: Infinite reconciliation loop in StatefulSet never recovering

dimpavloff:
**Description**

Resizing STS VolumeClaimTemplates is currently broken because the change does not get detected and acted on. This is an unintended side effect of #1322 where the `newStateful` has its VCT changes reset and then a patch is calculated. If you've only changed the VCT, then because of the reset no changes are detected and the function returns early, leading to not resizing the PVCs, not replacing the StatefulSet (if recreation is enabled), and not setting the storageCapacity annotation. 

The change now takes into account whether STS recreation is enabled or not. If it is, it makes sense to not reset the VCT and recreate the STS. This then solves the TODO of handling the case when the STS is scaled out -- the new PVCs will have the intended new size.

The change is slightly messy -- it resizes the PVCs first and only then checks if the StatefulSet should be patched; this logically seems backwards. However, I opted for this because the alternative would have required a larger refactor (e.g. extracting some of the logic from HandlePVCResizing which detects there is a change). 
Another way to go about this is to remove the logic resetting the VCT changes altogether; then the code becomes a lot simpler. Given that the operator supports re-creating the STS (including orphaning the resources which will be adopted by the new STS), I am struggling to see why re-create isn't always turned on. Nevertheless, I have opted to keep the behaviour as I could be missing something. 

Lastly, I have reviewed #1321 and I think my PR still addresses it. I have tested locally by mutating a RedisCluster's annotation, observing the operator updating the STS, the pods being rolled, and then the stream of changes from `kubectl get sts -w` stopping. 

Fixes #1321

**Type of change**

**Checklist**

**Additional Context**

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -231,6 +231,13 @@ jobs:
           config: tests/_config/kind-config.yaml
           cluster_name: kind
 
+      # NOTE: This is a workaround for the issue where the default storage class does not support volume expansion.
+      # Since we don't require PVC resizing (unlike physical disks), we can simply ensure that the requested PVC size is met.
+      - name: Set allowVolumeExpansion to true
+        run: |
+          DEFAULT_SC=$(kubectl get storageclass -o=jsonpath='{.items[?(@.metadata.annotations.storageclass\.kubernetes\.io/is-default-class=="true")].metadata.name}')
+          kubectl patch storageclass $DEFAULT_SC -p '{"allowVolumeExpansion": true}'          
+
       - name: Load Docker image into Kind
         run: |
           kubectl cluster-info --context kind-kind
--- a/internal/k8sutils/pvc.go
+++ b/internal/k8sutils/pvc.go
@@ -62,10 +62,6 @@ func HandlePVCResizing(ctx context.Context, storedStateful, newStateful *appsv1.
 
 	// If the stored capacity matches the desired capacity, no update is needed.
 	if storedCapacity == desiredCapacity {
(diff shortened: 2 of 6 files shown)


shubham-cmyk:
@dimpavloff 

Fix the conflict please and sign off commit

dimpavloff:
@shubham-cmyk Done. Please lmk if there's anything else I should do

dimpavloff:
I'm not sure why the lint is failing on GH. Locally `make fmt vet` runs without changes

drivebyer:
I would try to add an e2e test case to it.

shubham-cmyk:
@drivebyer 

I remember that kind didn't support the Volume scaling. 
Last time it worked on cloud as you confirmed.

drivebyer:
> @drivebyer
> 
> I remember that kind didn't support the Volume scaling. Last time it worked on cloud as you confirmed.

We can just assert the size in spec not the size in status

drivebyer (approved):

Issue #1429, opened 2025-07-03
Sentinel not being configured after pod created after deletion


solarisfire:
**What version of redis operator are you using?**

redis-operator version: v0.21.0

**Does this issue reproduce with the latest release?**

Yes

**What operating system and processor architecture are you using (`kubectl version`)?**

Talos OS 1.9.6
Kubernetes 1.32.6

**What did you do?**

I have a 3 pod redis replica setup, with 3 pods of sentinel.

When deleting the redis sentinel pods, they come back up, but do not get configured by the operator to point back to redis replication pods.

![Image](https://github.com/user-attachments/assets/d2298212-9a24-4cb8-b344-663a17a82263)

With no config in their /etc/redis/sentinel.conf:

![Image](https://github.com/user-attachments/assets/11e5ab8c-50f0-403c-9523-c97c781c81e1)

**What did you expect to see?**

Before v0.21.0 the sentinels would have been restarted and reconfigured one by one.

However this behavior was changed by: https://github.com/OT-CONTAINER-KIT/redis-operator/pull/1381

I would still, however, expect a restarted sentinel pod to be configured to point to the right master redis node.

**What did you see instead?**

See above. Sentinel remained unconfigured, whether deleting one pod or all three.

solarisfire:
If I uninstall the helm chart for the sentinel group, and install it again, things recover...

![Image](https://github.com/user-attachments/assets/1520db49-9196-4a51-9ca1-4f18c489f476)

However, this should be automated. These pods should recover on their own.

solarisfire:
And after deleting a sentinel pod again, it still does not get configured after many minutes.

drivebyer:
Maybe we need to watch the statefulset in redis sentinel controller.

solarisfire:
Reconciling periodically does fix it...

https://github.com/OT-CONTAINER-KIT/redis-operator/pull/1430/commits/21aae268d3a7e2caa87368df4b65a02a6524385c

Or yeah, could watch the statefulset.

drivebyer:
> Reconciling periodically does fix it...
> 
> [21aae26](https://github.com/OT-CONTAINER-KIT/redis-operator/commit/21aae268d3a7e2caa87368df4b65a02a6524385c)
> 
> Or yeah, could watch the statefulset.

I saw you raise a PR. The PR contain many changes include the Reconciling periodically change. I suggest to split it into multi PR for better track :)

[...]

Issue #1403, opened 2025-06-12
Redis Replication left with no master after failover


jzkinetic:
**What version of redis operator are you using?**

```shell
kubectl logs <_redis-operator_pod_name> -n <namespace>
{"level":"info","ts":"2025-06-12T00:50:29Z","msg":"Creating redis replication by executing replication creation commands","controller":"redisreplication","controllerGroup":"redis.redis.opstreelabs.in","controllerKind":"RedisReplication","RedisReplication":{"name":"dsv2-redis-replication-dev","namespace":"dev"},"namespace":"dev","name":"dsv2-redis-replication-dev","reconcileID":"8790ff5a-9682-4b7b-9dbd-b4cbdd1e2a65"}
[...]

PR #1430, opened 2025-07-03
Fix Redis master promotion and replication logic to prevent stale pod promotion

solarisfire:
**Description**

This PR improves the RedisReplication controller to correctly promote the replica with the most recent data when all other nodes are unavailable or in a broken state. It prevents a newly started empty Redis pod from being incorrectly promoted to master due to lack of existing masters.

Additionally, it ensures that once a healthy master is selected, all other pods are reattached as replicas and role labels are updated immediately within the same reconcile loop. This prevents having multiple pods temporarily labeled as `redis-role=master`, which could lead to misrouting or alerting issues.

Fixes [#1429](https://github.com/OT-CONTAINER-KIT/redis-operator/issues/1429) and [#1403](https://github.com/OT-CONTAINER-KIT/redis-operator/issues/1403)

**Type of change**

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

**Checklist**

**Additional Context**

This PR introduces a more robust fallback in the `reconcileRedis` function when no master is available by:
- Selecting the pod with the highest `master_repl_offset` as the new master.
- Ensuring stale or rebooted pods do not automatically become master just because they start first.
- Immediately updating `redis-role` labels to reflect the new topology, preventing dual-master labelling between reconcile loops.

This addresses production issues seen with sentinel reconcile race conditions and replication gaps on pod restart.

### DIFF
--- a/internal/controller/redisreplication/redisreplication_controller.go
+++ b/internal/controller/redisreplication/redisreplication_controller.go
@@ -2,6 +2,9 @@ package redisreplication
 
 import (
 	"context"
+	"fmt"
+	"strconv"
+	"strings"
 	"time"
 
 	rrvb2 "github.com/OT-CONTAINER-KIT/redis-operator/api/redisreplication/v1beta2"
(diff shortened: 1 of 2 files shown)


drivebyer:
I suggest to split this pr into multiple for better track, each pr link to issue it addressed :)

PR #1516, opened 2025-09-05
fix: Init-config fixes and update default versions

### LINKED ISSUE #1512: Init-conf on cluster getting hostname error
I see the following error for the init-config of each cluster replicas and neither the maxMemoryPercentOfLimit or additionalRedisConfg features are working 

`2025/09/04 14:42:36 Warning: Failed to get pod IP: exec: "hostname": executable file not found in $PATH`

chris-gooch:
**Description**

Fixes #1512 as latest tag was being used and older image was getting cached on node causing the error.

It also includes: 

1. `include /etc/redis/external.conf.d/*.conf` by default as the current logic does not work for init-config due to volume not being mounted to the pod at this stage. If no files exist this is just ignored.
2. `pidfile /etc/redis/redis.pid` to clear the permission denied while redis tries to write the PID file.
3. updates the default redis to v7.4.5 and exporter to v.1.76.0

**Type of change**

**Checklist**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,12 @@
 # Build the manager or agent binary
 FROM golang:1.23-alpine3.21 AS builder
+ARG IMG
+		RUN if [ -z "$IMG" ]; then \
+			echo "\n[ERROR] The required build argument IMG is missing!" >&2; \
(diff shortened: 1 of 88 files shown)


drivebyer:
fixed in https://github.com/OT-CONTAINER-KIT/redis-operator/pull/1530

PR #1537, opened 2025-09-12
fix: ServiceMonitor defaults to release.Namespace instead of monitoring namespace (#1507)

### LINKED ISSUE #1507: Improve ServiceMonitor
Hi,

https://github.com/OT-CONTAINER-KIT/redis-operator/blob/main/charts/redis-sentinel/values.yaml#L68
https://github.com/OT-CONTAINER-KIT/redis-operator/blob/main/charts/redis-sentinel/templates/servicemonitor.yaml#L29

i'm not sure why you set the namespaceSelector to monitoring by default: 

```bash
  namespaceSelector:
    matchNames:
      - monitoring
```

I know that a ServiceMonitor can be in another namespace than the service itself ... but the way you have done it it would not work out of the box because the helm chart and all services are installed in `ot-operators` and the ServiceMonitor only watches objects in the `monitoring` namespace? 

You should just change it to `{}` as default which is the same namespace like where the ServiceMonitor is installed to. 

Thanks,
Regards
Kilian

mjtrangoni:
**Description**

Hi,

This PR addresses an issue in the `redis-sentinel` Helm chart where the *ServiceMonitor* was previously configured with the hard-coded namespace monitoring.
The update modifies this behavior to reference the current `.Release.Namespace` instead. If a namespace is explicitly specified by the user, the *ServiceMonitor* will now be installed in the appropriate namespace.
  

Fixes #1507

**Type of change**

**Checklist**

**Additional Context**

### DIFF
--- a/charts/redis/README.md
+++ b/charts/redis/README.md
@@ -91,7 +91,7 @@ helm delete <my-release> --namespace <namespace>
 | serviceMonitor.enabled | bool | `false` |  |
 | serviceMonitor.extraLabels | object | `{}` | extraLabels are added to the servicemonitor when enabled set to true |
 | serviceMonitor.interval | string | `"30s"` |  |
-| serviceMonitor.namespace | string | `"monitoring"` |  |
+| serviceMonitor.namespace | string | `""` | Namespace where servicemonitor resource will be created, if empty it will be created in the same namespace as the redis |
 | serviceMonitor.scrapeTimeout | string | `"10s"` |  |
 | sidecars.env | list | `[]` |  |
 | sidecars.image | string | `""` |  |
--- a/charts/redis/templates/servicemonitor.yaml
+++ b/charts/redis/templates/servicemonitor.yaml
@@ -4,6 +4,7 @@ apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
   name: {{ .Values.redisStandalone.name | default .Release.Name }}-prometheus-monitoring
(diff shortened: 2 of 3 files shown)


mjtrangoni:
I was taking a closer look at the issue, and IMHO the object that can be installed somewhere else is the **ServiceMonitor**, not the k8s **Service** itself. So this fix the issue accordingly

See,
```bash
$ kubectl explain servicemonitor.metadata.namespace
GROUP:      monitoring.coreos.com
KIND:       ServiceMonitor
VERSION:    v1

FIELD: namespace <string>


DESCRIPTION:
    Namespace defines the space within which each name must be unique. An empty
    namespace is equivalent to the "default" namespace, but "default" is the
    canonical representation. Not all objects are required to be scoped to a
    namespace - the value of this field for those objects will be empty.

    Must be a DNS_LABEL. Cannot be updated. More info:
    https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces

$ kubectl explain servicemonitor.spec.namespaceSelector.matchNames
GROUP:      monitoring.coreos.com
KIND:       ServiceMonitor
VERSION:    v1

FIELD: matchNames <[]string>


DESCRIPTION:
    List of namespace names to select from.
```

@drivebyer let me know if I should fix all **ServiceMonitors** around, as we have here the very same problem everywhere.

Thank you!

drivebyer:
@mjtrangoni  It's a nice fix, we should fix in other charts as well

mjtrangoni:
@drivebyer perfect! I modified the other helm charts as you requested.

PR #1581, opened 2025-11-02
fix: :bug: Fix faulty CRD when no options are set

Supporterino:
**Type of change**


drivebyer:
could you please provide steps to reproduce it?

PR #1583, opened 2025-11-03
fix: Reload RedisCluster instance after status updates

naimadswdn:
**Description**

Fix race condition where status checks used stale instance data, preventing health checks from running. Reload instance after each status update to ensure subsequent checks have fresh status values.

This fixes cluster state reporting issues where ReadyLeaderReplicas and ReadyFollowerReplicas checks always failed due to stale data.

**Type of change**

**Checklist**

- [X] Tests have been added/modified and all tests pass.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -260,7 +260,7 @@ jobs:
       - name: Install Redis Operator
         run: |
           make deploy IMG=redis-operator:e2e
-          kubectl set env deployment/redis-operator-redis-operator -n redis-operator-system OPERATOR_IMAGE=redis-operator:e2e FEATURE_GATES=GenerateConfigInInitContainer=true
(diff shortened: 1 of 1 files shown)


naimadswdn:
@drivebyer I think I finally found the reason why Redis Cluster formation was such problematic (especially in the E2E tests). What do you think?

drivebyer:
> @drivebyer I think I finally found the reason why Redis Cluster formation was such problematic (especially in the E2E tests). What do you think?

~~We could reutrn after update status, the new reconcile loop will get the fresh status.~~

Maybe the update api conflict because stale CR, we could do retry on update.

drivebyer (approved):

drivebyer:
This approach doesn’t fix the flaky test. Maybe we need to try another way, such as returning early after updating the status.

drivebyer:
return early after update the status https://github.com/OT-CONTAINER-KIT/redis-operator/pull/1589

PR #1629, opened 2025-12-30
fix: RedisCluster - make scale-out resilient by waiting for node convergence and blocking rebalance/add-node while slots are open

simonnorra-sap:
**Description**

The RedisCluster controller can hit a non-idempotent / flaky scale-out failure during the v1beta2 chainsaw e2e test. The operator runs redis-cli --cluster rebalance ... --cluster-use-empty-masters and often gets:

CLUSTERDOWN The cluster is down mid clusterManagerMoveSlot

leaving open slots (migrating/importing, e.g. slot 5491)

subsequent reconciles then fail on add-node / rebalance because Redis refuses cluster management while slots are open

This shows up in the e2e flow (see setup-redis-cluster test) that does failover and then scale out shortly after.

**Root cause (what we believe is happening)**

The controller currently treats “pods are Ready / cluster_state ok on a seed” as sufficient to proceed, but during membership changes (failover + scale-out), a newly added empty master (or a node targeted for slot moves) can still be not fully converged from the cluster’s perspective.

Result: redis-cli --cluster rebalance starts slot migration to a node that still rejects cluster management commands with CLUSTERDOWN, which aborts the rebalance and may leave a slot in migrating/importing state. After that, the operator immediately continues with more cluster commands (rebalance/add-node) without first ensuring the cluster has no open slots.

**Fix summary**

Make scale-out reconcile converge-first, and make the reconcile loop idempotent in the presence of transient CLUSTERDOWN / partial slot-moves

1. Add waitForNodePresence gating
    * After add-node and before any rebalance, wait until:
        * the new node appears in CLUSTER NODES from the seed and is connected
        * the new node itself reports cluster_state:ok (via CLUSTER INFO)
        * cluster_known_nodes matches expected size for the current phase
2. Add an “open slot guard” before cluster-management commands
    * Before running rebalance, check the cluster for open slots (migrating/importing).
    * If open slots exist: do not proceed; requeue with a short backoff, log a clear message.

**Why this prevents the issue “anymore”**
This changes the operator from “issue cluster command ASAP” to “issue cluster command only when the cluster is in a safe state”:
* No rebalance until the target node is fully recognized and stable
* No add-node/rebalance while there are open slots
That makes the reconcile loop safe to repeat and resilient to temporary cluster instability caused by failover + rapid scale-out.

**Type of change**

### DIFF
--- a/tests/data-assert/resources.yaml
+++ b/tests/data-assert/resources.yaml
@@ -288,10 +288,10 @@ spec:
       resources:
         limits:
           cpu: "500m"
-          memory: "1Gi"
+          memory: "512Mi"
         requests:
           cpu: "100m"
-          memory: "256Mi"
+          memory: "128Mi"
       volumeMounts:
         - name: data-assert
           mountPath: /go/src/data-assert


simonnorra-sap:
Hey @drivebyer and team

whenever you have a moment, could you please take a look and share your feedback on this PR?

Thanks in advance!

drivebyer:
> Hey @drivebyer and team
> 
> whenever you have a moment, could you please take a look and share your feedback on this PR?
> 
> Thanks in advance!

Sorry for the late. It's a nice PR! I will review this by the weekend :)

drivebyer (approved):
/lgtm

PR #1647, opened 2026-01-28
fix: Make RedisCluster scaling resilient to failover and open slots

shubham-cmyk:
## Summary

  - **Failover safety during scale-down:** Before resharding during scale-down, the controller now verifies that all remaining leader pods are actual masters. If a previous scale-out triggered a failover that demoted a leader pod to a replica, the controller initiates a `CLUSTER FAILOVER` and requeues, preventing the "specified node is not known or not a master" reshard error.

  - **Fix open slots before scale-up:** Adds a new `FixRedisCluster()` helper that runs `redis-cli --cluster fix` to resolve migrating/importing slots left over from a previously interrupted rebalance. This is called before `add-node` / `rebalance` during scale-up so the cluster is in a clean state.

  - **Fix Sentinel first-time monitor setup:** `SentinelMonitor` now skips the `SENTINEL REMOVE` call when the master group doesn't exist yet, preventing an "ERR No such master with that name" error on initial setup.

  - **Typo fix:** Corrects "inteface" → "interface" in Storage type descriptions across CRDs and API types.

### DIFF
--- a/internal/controller/rediscluster/rediscluster_controller.go
+++ b/internal/controller/rediscluster/rediscluster_controller.go
@@ -94,6 +94,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 			// Before resharding, ensure all remaining leader pods (the transfer targets) are masters.
 			// After scale-out, a failover may have converted some leader pods to slaves, which causes
 			// reshard to fail with "The specified node is not known or not a master".
+			// We handle one failover per reconcile cycle and requeue — the loop will converge
+			// over successive reconciliations until all target pods are masters.
 			for i := int32(0); i < leaderReplicas; i++ {
 				if !(k8sutils.VerifyLeaderPod(ctx, r.K8sClient, instance, i)) {
 					logger.Info("Transfer target leader pod is not a master, initiating failover before scale-down", "Pod.Index", i)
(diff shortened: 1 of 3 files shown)


shubham-cmyk:
@drivebyer There was a bug in the setup/redis-cluster which is fixed now in this PR.

shubham-cmyk:
@drivebyer the change you have pushed is failing the CI should we remove this

drivebyer:
> @drivebyer the change you have pushed is failing the CI should we remove this

I think it introduced by https://github.com/OT-CONTAINER-KIT/redis-operator/pull/1646 in action https://github.com/OT-CONTAINER-KIT/redis-operator/actions/runs/21478189152/job/61868540450

shubham-cmyk:
@drivebyer i reviewed that PR that was fine, at that time I created this PR to check if the CI was failing on main or not. 
It in deed was failing. 

Let's try removing the change you did and CI would pass.

drivebyer:
BTW, the flaky scale-out failure already fixed by https://github.com/OT-CONTAINER-KIT/redis-operator/pull/1629. I am not sure this PR is address the same issue.

shubham-cmyk:
Or let me force push it wait

shubham-cmyk:
@drivebyer I have forced push let's see if it passes

drivebyer:
Maybe another flaky test caused by https://github.com/OT-CONTAINER-KIT/redis-operator/pull/1646

shubham-cmyk:
@drivebyer CI passed

drivebyer:
> @drivebyer CI passed

We should make CI passed on latest main branch

shubham-cmyk:
@drivebyer i  think this PR https://github.com/OT-CONTAINER-KIT/redis-operator/pull/1629
has handled correctly right way for the open slot's but it does after adding the node,  mine change is the just an extra check just before the adding new node if we need to make sure cluster don't have any open slot. 

i would try some e2e test on this to check if cluster goes down or not while scaling up.

PR #1653, opened 2026-02-02
fix: do not create empty PVC when storage has only volumeMount

aa1ex:
**Description**

When `spec.storage` contains only `volumeMount` entries (e.g. `emptyDir`) without `volumeClaimTemplate`, the operator incorrectly set `PersistenceEnabled=true`, causing an empty `volumeClaimTemplates` entry in the resulting StatefulSet. This led to unnecessary (and possibly broken) PVC provisioning.

**Root cause:** `PersistenceEnabled` was set based solely on `cr.Spec.Storage != nil`, without checking whether `volumeClaimTemplate` is actually populated. Since the `Storage` struct is non-nil whenever any storage is configured (including volumeMount-only with emptyDir), the operator always treated it as a request for a PersistentVolumeClaim.

**Changes:**
- Added `storageHasVolumeClaimTemplate()` helper in `statefulset.go` that checks whether `VolumeClaimTemplate` has meaningful content (`AccessModes`, `Resources.Requests`, `StorageClassName`, or `VolumeName`)
- Redis standalone and replication: replaced `cr.Spec.Storage != nil` with `storageHasVolumeClaimTemplate()` for both container and init container params
- Redis cluster init container: aligned the check with the existing container params logic (`PersistenceEnabled != nil && *PersistenceEnabled`)
- Added unit tests for `storageHasVolumeClaimTemplate`

**Type of change**

**Checklist**

**Additional Context**

Users who have `volumeClaimTemplate` properly configured are not affected — behavior is unchanged for them. The fix only removes the incorrect empty PVC creation for volumeMount-only storage configurations.

### DIFF
--- a/internal/k8sutils/redis-replication.go
+++ b/internal/k8sutils/redis-replication.go
@@ -189,7 +189,7 @@ func generateRedisReplicationContainerParams(cr *rrvb2.RedisReplication) contain
 	if cr.Spec.LivenessProbe != nil {
 		containerProp.LivenessProbe = cr.Spec.LivenessProbe
 	}
-	if cr.Spec.Storage != nil {
+	if storageHasVolumeClaimTemplate(cr.Spec.Storage) {
 		containerProp.PersistenceEnabled = &trueProperty
 	}
 	if cr.Spec.TLS != nil {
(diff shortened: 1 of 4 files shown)


shubham-cmyk (changes requested):
e2e test are failing ptal

PR #1660, opened 2026-02-05
chore: update version to 0.24.0

naimadswdn:
**Description**
Motivated by the recent changes related to webhook and the ACL setup from the PVC.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 # ===========================
 
 # Current Operator version
-VERSION ?= 0.23.0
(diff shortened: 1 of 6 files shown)


naimadswdn:
@shubham-cmyk @drivebyer @iamabhishek-dubey how about pushing new version of the operator? Can you please help me with that by approving that PR? If there is anything else to do, I can help as well

shubham-cmyk:
@naimadswdn Let's release this in march

PR #1694, opened 2026-02-27
feat: add preStop lifecycle hook for replication mode

peins94:
**Description**

  Replication mode pods currently have no preStop lifecycle hook, unlike cluster mode.
  When a master pod is terminated during a rolling update or node drain, clients experience
  downtime until Sentinel detects the failure and triggers a failover (up to `down-after-milliseconds`).
  This adds a preStop hook for the `replication` role following the same pattern as the existing
  cluster mode hook in `generateClusterPreStop`. The hook:
  1. Checks if the terminating pod is the master via `INFO replication`
  2. Derives the Sentinel headless service from the pod hostname
  3. Triggers `SENTINEL FAILOVER mymaster`
  4. Waits up to 30 seconds for the role to change to slave
  Replica pods skip the hook entirely.

  **Type of change**

  **Checklist**

  **Additional Context**

  In production we observed that every rolling restart of a RedisReplication StatefulSet causes
  client downtime because the master pod is killed without failover. This is particularly
  impactful for large datasets where replica startup takes 30-40 minutes, as Sentinel may
  trigger unnecessary full resyncs while the topology is unstable.

### DIFF
--- a/api/redisreplication/v1beta2/redisreplication_types_helper.go
+++ b/api/redisreplication/v1beta2/redisreplication_types_helper.go
@@ -18,6 +18,14 @@ func (cr *RedisReplication) SentinelHLService() string {
 	return cr.Name + "-s-hl"
 }
 
+// SentinelMasterName returns the master group name monitored by the embedded
+// Sentinel. The embedded Sentinel always monitors a single group named
+// "mymaster" (see internal/agent/bootstrap/sentinel/config.go, which falls back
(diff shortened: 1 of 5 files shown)


shubham-cmyk (approved):

peins94:
Can we re-run the e2e test please? I believe there might be some flakiness in the test itself as there were no code changes to the RedisCluster logic

PR #1696, opened 2026-02-27
fix: race condition with global serviceType variable

JulianPaulus:
**Description**
We had issues in our clusters where sometimes for RedisReplication the additional service would temporarily get changed to `ClusterIP` and then back to what it's supposed to be in our case: `Loadbalancer`. We do have MAX_CONCURRENT_RECONCILES set to something > 1 and noticed that the occurrence of this bug decreases with lower concurrency settings. Turns out the function to map strings to ServiceTypes uses a global variable that introduced a race condition between concurrent reconciles creating services during the same timeframes. This PR aims to fix this issue.

**Type of change**

**Checklist**


shubham-cmyk (approved):
LGTM!

Thanks for the PR

PR #1698, opened 2026-03-02
fix: fall back to system trust store when CA cert is absent from TLS …

### LINKED ISSUE #1697: TLS: operator should fall back to system trust store when CA cert is absent from secret
## Problem

When TLS is enabled, the operator requires `ca.crt` (or the overridden CA key name) to be present in the TLS secret. If it's missing, `getRedisTLSConfig` returns `nil` and the operator fails to connect to Redis entirely.

This is incorrect when Redis uses a certificate signed by a publicly trusted CA (e.g., Let's Encrypt, Google Trust Services). In that case, the operator's container image already has the CA in its system trust store, and no explicit `ca.crt` in the secret is needed.

### Affected code paths

- `internal/k8sutils/secrets.go` — `getRedisTLSConfig` returns `nil` if CA key is missing from secret
- `internal/controller/common/redis/heal.go` — same behavior in the duplicate `getRedisTLSConfig`
- `internal/k8sutils/redis.go` — `getRedisTLSArgs` always passes `--cacert /tls/<ca>` to `redis-cli`, which fails if the file doesn't exist on the pod

### Current behavior

```go
if !certExists || !keyExists || !caExists {
// returns nil — operator cannot talk to Redis at all
return nil
}
```

### Expected behavior

- If the CA key is absent from the secret **and** the user has not explicitly configured a CA filename, the operator should fall back to the system certificate pool (i.e., set `RootCAs: nil` in `tls.Config`, which makes Go use the system trust store).
- If the user explicitly configured a CA filename and it's missing from the secret, that is a configuration error and should still fail with a clear error.
- For `redis-cli` args, `--cacert` should be omitted when no CA file is mounted, so Redis CLI uses its own default trust behavior.

### Scenario matrix

| CA in secret | CA field set in TLSConfig | Expected behavior |
|---|---|---|
| Present | Yes or default | Use CA from secret (current behavior, works) |
| Absent | Explicitly set by user | Fail with clear error (user misconfiguration) |
| Absent | Not set (empty/default) | Fall back to system trust store |

asaxena7-atlas:
## Description

When using TLS with publicly trusted certificates (e.g. Let's Encrypt, Google Trust Services), users no longer need to supply a `ca.crt` in their TLS secret. The operator’s container image already trusts these CAs via its system trust store. Previously, the operator would fail if `ca.crt` was missing from the secret, even when it wasn’t needed.

This change makes the CA certificate optional:

- If `CaKeyFile` is **not** explicitly set in the CR and `ca.crt` is absent from the secret, the operator falls back to the system trust store (`RootCAs: nil` in Go’s `tls.Config`).
- If `CaKeyFile` **is** explicitly set but the corresponding key is missing from the secret, the operator fails fast with a clear error (treating it as misconfiguration).
- `redis-cli` commands, health probes, and exporter env vars conditionally omit `--cacert` / `REDIS_EXPORTER_TLS_CA_CERT_FILE` when no explicit CA is configured, preventing references to non-existent files.

Fixes #1697.

---

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

---

## Checklist

---

## Additional context

**Files changed:**

- `internal/k8sutils/secrets.go`  
  - `getRedisTLSConfig` now returns a `tls.Config` with `RootCAs = nil` (system trust) when the CA is absent and not explicitly configured.

- `internal/controller/common/redis/heal.go`  
  - Applies the same CA-optional logic in the duplicated `getRedisTLSConfig`.

- `internal/k8sutils/redis.go`  
  - `getRedisTLSArgs` omits `--cacert` when `CaKeyFile` is empty.

- `internal/k8sutils/statefulset.go`  
  - `GenerateTLSEnvironmentVariables` and `getExporterEnvironmentVariables` conditionally emit CA-related env vars.
  - `GenerateAuthAndTLSArgs` and health probes now use shell parameter expansion (`${REDIS_TLS_CA_KEY:+--cacert ...}`) to conditionally include `--cacert`.

- **Tests**
  - `secrets_test.go`, `redis_test.go`, and `statefulset_test.go` have been updated to cover absent-CA scenarios.

This change is fully backward compatible: existing setups that include `ca.crt` in their secret continue to work identically.

### DIFF
--- a/internal/agent/bootstrap/redis/config.go
+++ b/internal/agent/bootstrap/redis/config.go
@@ -108,7 +108,11 @@ func GenerateConfig() error {
 	if tlsMode == "true" {
 		cfg.Append("tls-cert-file", util.CoalesceEnv1("REDIS_TLS_CERT", ""))
 		cfg.Append("tls-key-file", util.CoalesceEnv1("REDIS_TLS_CERT_KEY", ""))
-		cfg.Append("tls-ca-cert-file", util.CoalesceEnv1("REDIS_TLS_CA_CERT", ""))
+		if caCert := util.CoalesceEnv1("REDIS_TLS_CA_CERT", ""); caCert != "" {
+			cfg.Append("tls-ca-cert-file", caCert)
+		} else {
+			cfg.Append("tls-ca-cert-file", "/etc/ssl/certs/ca-certificates.crt")
+		}
 		cfg.Append("tls-auth-clients", "optional")
 		cfg.Append("tls-replication", "yes")
 
--- a/internal/agent/bootstrap/redis/config.go
+++ b/internal/agent/bootstrap/redis/config.go
@@ -32,7 +32,8 @@ pidfile /var/run/redis.pid
 
 // GenerateConfig generates Redis configuration file
 func GenerateConfig() error {
-	cfg := agentutil.NewConfig("/etc/redis/redis.conf", defaultRedisConfig)
+	confPath := util.CoalesceEnv1("REDIS_CONFIG_FILE", "/etc/redis/redis.conf")
+	cfg := agentutil.NewConfig(confPath, defaultRedisConfig)
 
 	var (
 		persistenceEnabled = util.CoalesceEnv1("PERSISTENCE_ENABLED", "false")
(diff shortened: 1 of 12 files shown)


artkoshelev:
Hey folks, can some take a look at this PR? @shubham-cmyk @drivebyer @iamabhishek-dubey

shubham-cmyk (approved):

Issue #1407, opened 2025-06-17
RedisCluster node cannot init/rejoin after cluster scale-in and scale-out due to already stored configuration


dimpavloff:
When you scale in a RedisCluster with nodeConf volumes, these volumes are preserved. When you subsequently scale out a RedisCluster, the "new" pods re-use the PVCs but because of this, the command to add them to the cluster fails. 

A manual workaround has been to delete the nodeConf PVCs.
Another workaround is to not use the nodeConf persistence but this breaks in other ways -- restarting a pod means it it will not rejoin the existing cluster automatically and instead the operator will need to handle it (it doesn't). Also, if leader-0 restarts, the operator treats this as needing to initialise a new cluster, which fails because the other nodes are already with data. 

A solution could be to allow setting and deleting the volumes on scale-in [PVC Claim Policy]( https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#persistentvolumeclaim-retention). This is probably fine although it means that if you temporary scale down the StatefulSet (e.g. to save costs; it requires disabling reconciliation), you will lose the data. This is a niche use case, though, so I think this is probably a sensible approach. Another could be for the Redis operator to handle this during cluster scale in  -- it could delete the PVCs when node has successfully been removed [here](https://github.com/OT-CONTAINER-KIT/redis-operator/blob/main/internal/controller/rediscluster/rediscluster_controller.go#L106) (or perhaps run FLUSHALL + CLUSTER RESET HARD ). What do you think?

**What version of redis operator are you using?**
redis-operator version: v0.20.2

**Does this issue reproduce with the latest release?**
Yes

**What operating system and processor architecture are you using (`kubectl version`)?**
x86, K8s 1.26 (yes, this is old)

**What did you do?**

1. Create a RedisCluster with nodeConf volume and 10 leaders
1. Scale in the cluster to 8 
1. Wait for scale in to complete (`redis-cli --cluster check localhost:6379` returns all green)
1. Scale out the cluster back to 10 (Note that the PVCs for pods `reds-cluster-leader-8` and `reds-cluster-leader-9` already exist)
1. Observe the bug:
  * redis-operator is logging errors, continuously [1] 
  * logs from reds-cluster-leader-8 and 9 do not show it successfully joining [2]
  * the number of reported masters from `redis-cli --cluster check localhost:6379` remains 8, not 10 
[...]

PR #1705, opened 2026-03-09
fix: RepairDisconnectedMasters does not heal failed followers, causing infinite Failed state after pod restarts

### LINKED ISSUE #1692: RepairDisconnectedMasters does not heal failed followers, causing infinite Failed state after pod restarts

redis-operator version: v0.23.0 (chart 0.23.0, image `quay.io/opstree/redis-operator:v0.23.0`)

Yes, v0.23.0 is the latest. The bug has been present for a while — #1179 reported the same thing back in Dec 2024 on v0.18.1.

$ kubectl version
Client Version: v1.35.0
Kustomize Version: v5.7.1
Server Version: v1.35.0-33+37970203ae1a44
</pre></details>

**What did you do?**

Running a RedisCluster (`clusterSize: 3`, 3 leaders + 3 followers) on EKS. Initial deployment works perfectly fine. The problem shows up after any pod restart — node rotation, drain, spot reclaim, rollout restart, etc. Once follower pods come back with new IPs, the cluster breaks and never self-heals.

The operator should automatically reconnect the followers after they get new IPs, just like it does for masters.

**What did you see instead?**

The operator gets stuck in an infinite `Failed` loop. Followers log:

```
Error condition on socket for SYNC: Host is unreachable
Connecting to MASTER <old-ip>:6379
```

Operator logs repeat forever:
```
"Number of Redis nodes match desired"
"healthy leader count does not match desired; attempting to repair disconnected masters"
```

CR status:
```yaml
status:
  state: Failed
  reason: "RedisCluster has unhealthy nodes"
```

The only way to recover is manual `CLUSTER MEET` + `CLUSTER REPLICATE` (see workaround below). This happens on every single pod restart for followers.

### Root Cause

Two things go wrong:

**1. `repairDisconnectedMasters` only repairs masters, skips slaves**

In `redis.go` lines 158-162:
```go
masterNodeType := "master"
for _, node := range nodes {
    if !nodeIsOfType(node, masterNodeType) {
        continue   // <-- skips all failed slaves
    }
```

When followers get new IPs, the leaders see the old follower IPs as `fail` in gossip. But the repair function only looks at failed masters — which are healthy — so it does nothing.

**2. `CheckRedisNodeCount` counts failed nodes toward the total**

In `redis.go` line 420:
```go
count := len(clusterNodes)
```

The failed followers are still in the `CLUSTER NODES` output (just with `fail` flag), so they get counted. This means `nc == totalReplicas` at `rediscluster_controller.go` line 231, and the controller skips the code path that has `ExecuteRedisReplicationCommand` — the only function that could reconnect followers.

**3. Even after fixing gossip, replication stays broken**

Running `CLUSTER MEET` from a leader to a follower's new IP fixes the gossip layer. But the follower's `REPLICAOF` is still pointed at the master's old IP. You also need to run `CLUSTER REPLICATE <master-node-id>` on the follower to make it re-resolve the master's current IP from gossip.

**4. The retry doesn't escalate**

The escalation check `unhealthyNodeCount >= int(totalReplicas)-1` (e.g. `3 >= 5`) is false for the common case (3 failed followers out of 6 total), so the "manual intervention required" error never fires. It just requeues every 10 seconds forever.

### Manual Workaround

```bash
# Step 1: Fix gossip — CLUSTER MEET from leader-0 to each follower's current IP
kubectl -n redis exec <leader-pod> -- redis-cli cluster meet <follower-0-new-ip> 6379
kubectl -n redis exec <leader-pod> -- redis-cli cluster meet <follower-1-new-ip> 6379
kubectl -n redis exec <leader-pod> -- redis-cli cluster meet <follower-2-new-ip> 6379

# Step 2: Fix replication — CLUSTER REPLICATE on each follower with its master's node ID
kubectl -n redis exec <follower-0> -- redis-cli cluster replicate <master-0-node-id>
kubectl -n redis exec <follower-1> -- redis-cli cluster replicate <master-1-node-id>
kubectl -n redis exec <follower-2> -- redis-cli cluster replicate <master-2-node-id>
```

### Suggested Fix

Broaden `repairDisconnectedMasters` to also handle failed slaves. The existing `getRedisServerIP` already resolves any pod's current IP from the Kubernetes API:

```go
func repairDisconnectedNodes(ctx context.Context, client kubernetes.Interface, cr *rcvb2.RedisCluster, redisClient *redis.Client) error {
    nodes, err := clusterNodes(ctx, redisClient)
    if err != nil {
        return err
    }
    var lastError error
    for _, node := range nodes {
        if !nodeFailedOrDisconnected(node) {
            continue
        }
        host, err := getMasterHostFromClusterNode(node)
        if err != nil {
            lastError = err
            continue
        }
        ip := getRedisServerIP(ctx, client, RedisDetails{
            PodName:   strings.Split(host, ".")[0],
            Namespace: cr.Namespace,
        })
        // Step 1: fix gossip
        if err = redisClient.ClusterMeet(ctx, ip, strconv.Itoa(*cr.Spec.Port)).Err(); err != nil {
            lastError = err
            continue
        }
        // Step 2: if slave, re-establish replication so it resolves the master's new IP
        if nodeIsOfType(node, "slave") {
            followerClient := configureRedisClient(ctx, client, cr, strings.Split(host, ".")[0])
            masterNodeID := node[3]
            if err = followerClient.ClusterReplicate(ctx, masterNodeID).Err(); err != nil {
                lastError = err
            }
            followerClient.Close()
        }
    }
    return lastError
}
```

Both `CLUSTER MEET` and `CLUSTER REPLICATE` are idempotent and safe to retry.

---

I've been running this operator in production across multiple EKS clusters and this keeps biting us on every node rotation. I'd be happy to put together a PR if the approach above looks reasonable to you.

svils:
## Summary

Fixes #1692

- Broaden `repairDisconnectedMasters` → `RepairDisconnectedNodes` to handle both failed masters **and** failed slaves. For slaves, issue `CLUSTER MEET` (fix gossip) + `CLUSTER REPLICATE` (re-establish replication so the follower resolves its master's current IP from gossip).
- Add `RepairStaleReplication` to detect connected followers whose `master_link_status` is `down` and re-issue `CLUSTER REPLICATE`. This catches the subtler case where gossip is fine but replication is still pointed at a stale IP.
- Harden cluster config defaults to reduce the likelihood and blast radius of the problem:
  - `tcp-keepalive` 60s (Redis recommends this; faster dead connection detection)
  - `cluster-node-timeout` 15000ms, configurable via `CLUSTER_NODE_TIMEOUT` env var (more gossip recovery time before marking nodes failed)
  - `cluster-allow-reads-when-down yes` (keep clients unblocked while the operator repairs nodes)
  - Probe `TimeoutSeconds=5` / `FailureThreshold=5` (prevent Kubernetes from prematurely evicting pods during transient blips)

Both `CLUSTER MEET` and `CLUSTER REPLICATE` are idempotent and safe to retry.

## Test plan

- [x] Existing tests updated for the rename (`RepairDisconnectedMasters` → `RepairDisconnectedNodes`)
- [x] New unit tests for `replicationLinkUp` (up, down, master node, empty)
- [x] New unit tests for `RepairStaleReplication` (replication down → repaired, replication up → no-op)
- [x] Updated `statefulset_test.go` expectations for new probe defaults
- [x] All tests pass (`go test ./internal/k8sutils/...`)
- [x] Deploy to a test cluster, restart follower pods, verify automatic recovery
- [x] Deploy to a test cluster, restart leader pods, verify followers re-resolve master IP

### DIFF
--- a/internal/agent/bootstrap/redis/config.go
+++ b/internal/agent/bootstrap/redis/config.go
@@ -17,11 +17,14 @@ import (
 )
 
 // defaultRedisConfig from https://github.com/OT-CONTAINER-KIT/redis/blob/master/redis.conf
+// tcp-keepalive is lowered from the Redis default (300s) to 60s so that dead
+// peer connections are detected faster, which helps the cluster gossip layer
+// converge sooner after pod restarts with new IPs.
 const defaultRedisConfig = `
 bind 0.0.0.0 ::
 tcp-backlog 511
 timeout 0
-tcp-keepalive 300
+tcp-keepalive 60
 daemonize no
 supervised no
 pidfile /var/run/redis.pid
@@ -57,11 +60,23 @@ func GenerateConfig() error {
 	if clusterMode == "cluster" {
 		nodeConfPath := filepath.Join(nodeConfDir, "nodes.conf")
 
+		// Cluster-mode tuning (see also tcp-keepalive above):
+		//  - cluster-node-timeout raised from 5000ms to 15000ms (configurable
+		//    via CLUSTER_NODE_TIMEOUT) to give gossip time to converge after
+		//    pod restarts before marking nodes as failed.
+		//  - cluster-allow-reads-when-down (and pubsubshard variant on v7)
+		//    keeps clients unblocked while the operator repairs nodes.
+		clusterNodeTimeout := util.CoalesceEnv1("CLUSTER_NODE_TIMEOUT", "15000")
+
 		cfg.Append("cluster-enabled", "yes")
-		cfg.Append("cluster-node-timeout", "5000")
+		cfg.Append("cluster-node-timeout", clusterNodeTimeout)
 		cfg.Append("cluster-require-full-coverage", "no")
 		cfg.Append("cluster-migration-barrier", "1")
 		cfg.Append("cluster-config-file", nodeConfPath)
+		cfg.Append("cluster-allow-reads-when-down", "yes")
+		if redisMajorVersion == "v7" {
+			cfg.Append("cluster-allow-pubsubshard-when-down", "yes")
+		}
 
 		if ip, err := util.GetLocalIP(); err != nil {
 			log.Printf("Warning: Failed to get local IP: %v", err)
--- a/internal/agent/bootstrap/redis/config.go
+++ b/internal/agent/bootstrap/redis/config.go
@@ -60,12 +60,9 @@ func GenerateConfig() error {
 	if clusterMode == "cluster" {
 		nodeConfPath := filepath.Join(nodeConfDir, "nodes.conf")
 
-		// Cluster-mode tuning (see also tcp-keepalive above):
-		//  - cluster-node-timeout raised from 5000ms to 15000ms (configurable
-		//    via CLUSTER_NODE_TIMEOUT) to give gossip time to converge after
-		//    pod restarts before marking nodes as failed.
-		//  - cluster-allow-reads-when-down (and pubsubshard variant on v7)
-		//    keeps clients unblocked while the operator repairs nodes.
+		// cluster-node-timeout is raised from 5000ms to 15000ms (configurable
+		// via CLUSTER_NODE_TIMEOUT) to give gossip time to converge after
+		// pod restarts before marking nodes as failed.
 		clusterNodeTimeout := util.CoalesceEnv1("CLUSTER_NODE_TIMEOUT", "15000")
 
 		cfg.Append("cluster-enabled", "yes")
@@ -73,10 +70,6 @@ func GenerateConfig() error {
 		cfg.Append("cluster-require-full-coverage", "no")
 		cfg.Append("cluster-migration-barrier", "1")
 		cfg.Append("cluster-config-file", nodeConfPath)
-		cfg.Append("cluster-allow-reads-when-down", "yes")
-		if redisMajorVersion == "v7" {
-			cfg.Append("cluster-allow-pubsubshard-when-down", "yes")
-		}
 
 		if ip, err := util.GetLocalIP(); err != nil {
 			log.Printf("Warning: Failed to get local IP: %v", err)
--- a/internal/controller/rediscluster/rediscluster_controller.go
+++ b/internal/controller/rediscluster/rediscluster_controller.go
@@ -346,9 +346,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 			return intctrlutil.Requeue()
 		}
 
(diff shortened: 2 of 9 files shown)


svils:
@shubham-cmyk hey, you asked me to open this PR back on #1692 — it's been about a month now with no review. we're running this in production across multiple EKS clusters and keep hitting these bugs on every node rotation.

this PR (#1705) fixes the follower repair loop, but while debugging in prod i ran into two more issues that are closely related:

- #1706 — single-leader clusters get stuck in Bootstrap forever because ADDSLOTS exceeds the k8s exec URL limit
- #1734 — after the leader-only CLUSTER RESET, `add-node` fails with "Node is not empty" because the follower still has stale state (same root cause as #1407)

all three are straightforward fixes with tests, and they've been running stable in our production builds. would be great to get some eyes on these so they can land in the next release.

joepizza1:
@shubham-cmyk @drivebyer can you please review and get this fix in ? We're facing this issue on few environmnets and it would be really helpful to have this fixed.

shubham-cmyk:
Pushed two commits to address the review feedback (thanks @svils for the fix — the core repair logic is good and is kept as-is):

- **Dropped the global probe defaults** (`TimeoutSeconds`/`FailureThreshold` = 5 in `getProbeInfo`): that helper is shared by readiness *and* liveness probes for all Redis kinds, so the new defaults would delay restarts of genuinely dead pods (~50s) operator-wide. Probe timings are already configurable per CR; if we want hardened defaults, that deserves its own discussed PR.
- **Dropped the hardcoded `cluster-allow-reads-when-down yes`** (and the pubsubshard variant): flipping the cluster's availability/consistency default is a semantic change that shouldn't ride along silently in a bugfix. Happy to consider it as a separate, configurable option.
- **Documented why `RepairStaleReplication` runs every reconcile**: a broken replication link is invisible to gossip (the follower still shows `connected` in `CLUSTER NODES`), so it can't be gated on the unhealthy-node count. Cost is one `CLUSTER NODES` call plus one `INFO replication` per connected follower, in line with the other per-reconcile checks.
- **Added unit coverage for the new slave-repair path** (failed follower → `CLUSTER MEET` + `CLUSTER REPLICATE` on the follower client) and the empty-IP skip branch, by making the follower client injectable in `repairDisconnectedNodes` — same pattern as `repairStaleReplication`.
- Merged `main` into the branch (it was behind) and fixed gofmt issues.

The `tcp-keepalive 60` and env-configurable `cluster-node-timeout` changes are kept.

shubham-cmyk (dismissed):

shubham-cmyk (approved):

PR #1732, opened 2026-04-01
fix(chart): make minAvailable and maxUnavailable optional in redis-replication chart

SvenMark:
**Description**

The poddisruptionbudget specification states that either `minAvailable` or `maxUnavailable` should be set.

When using server-side apply we get an error that `maxUnavailable` is set even though it is 'null' in the chart's values:
```bash
The RedisReplication "test" is invalid: spec.pdb.maxUnavailable: Invalid value: "null": spec.pdb.maxUnavailable in body must be of type integer: "null"
```

The provided change makes the `minAvailable` or `maxUnavailable` values optional. Either one of them could be set, but not both.

**Type of change**

**Checklist**

**Additional Context**

### DIFF
--- a/charts/redis-replication/Chart.yaml
+++ b/charts/redis-replication/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v2
 name: redis-replication
 description: Provides easy redis setup definitions for Kubernetes services, and deployment.
-version: 0.17.0
-appVersion: "0.17.0"
+version: 0.17.1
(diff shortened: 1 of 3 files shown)


charts/redis-replication/Chart.yaml
drivebyer, line 1: Revert this change.

drivebyer (changes requested):

PR #1749, opened 2026-04-21
fix(cluster): only render pdb.maxUnavailable / minAvailable when set

### LINKED ISSUE #1745: redis-cluster Helm chart renders both minAvailable and maxUnavailable in PDB causing RedisCluster reconciliation failure

```shell
redis-operator version: v0.20.2
redis-cluster chart version: 0.17.2
```

Yes. The issue exists in the current `_helpers.tpl` template in the redis-cluster Helm chart.

$ kubectl version
Client Version: v1.32.2
</pre></details>

**What did you do?**

Enabled PDB for a RedisCluster via the redis-cluster Helm chart with only `maxUnavailable` set:

```yaml
redisCluster:
  leader:
    pdb:
      enabled: true
      maxUnavailable: 1
  follower:
    pdb:
      enabled: true
      maxUnavailable: 1
```

The chart default `values.yaml` defines both `maxUnavailable: 1` and `minAvailable: 1` under `pdb`. The `_helpers.tpl` template at line 33-38 unconditionally renders both fields when PDB is enabled:

```yaml
{{- if .pdb.enabled  }}
pdb:
  enabled: {{ .pdb.enabled }}
  maxUnavailable: {{ .pdb.maxUnavailable }}
  minAvailable: {{ .pdb.minAvailable }}
{{- end }}
```

This causes the RedisCluster CR to always contain both `minAvailable` and `maxUnavailable` regardless of what the user configures. The operator then passes both values to the Kubernetes API when creating the PodDisruptionBudget.

The PDB should be created successfully with only `maxUnavailable: 1`. The template should conditionally render only the field that is set, so that users can choose either `minAvailable` or `maxUnavailable` without conflict.

**What did you see instead?**

The operator fails to create the PDB and the RedisCluster reconciliation is blocked with the following error:

```
Redis PodDisruptionBudget creation failed: PodDisruptionBudget.policy "cluster-native-cache-leader" is invalid: spec: Invalid value: minAvailable and maxUnavailable cannot be both set
```

This blocks the entire cluster initialization -- follower StatefulSets are never created because the reconciliation loop keeps failing on PDB creation.

**Suggested fix**

Update `charts/redis-cluster/templates/_helpers.tpl` to conditionally render PDB fields:

```yaml
{{- if .pdb.enabled }}
pdb:
  enabled: {{ .pdb.enabled }}
{{- if .pdb.maxUnavailable }}
  maxUnavailable: {{ .pdb.maxUnavailable }}
{{- end }}
{{- if .pdb.minAvailable }}
  minAvailable: {{ .pdb.minAvailable }}
{{- end }}
{{- end }}
```

And update `charts/redis-cluster/values.yaml` defaults to only set one:

```yaml
pdb:
  enabled: true
  maxUnavailable: 1
```

SAY-5:
## Summary

Fixes #1745.

The `_helpers.tpl` PDB block unconditionally rendered both `maxUnavailable` and `minAvailable` from `values.yaml`, which the chart defaults to `(1, 1)`:

```gotemplate
{{- if .pdb.enabled }}
pdb:
  enabled: {{ .pdb.enabled }}
  maxUnavailable: {{ .pdb.maxUnavailable }}
  minAvailable: {{ .pdb.minAvailable }}
{{- end }}
```

When a user enabled the PDB without clearing one of them, the RedisCluster CR carried both fields and the operator propagated them to the PodDisruptionBudget controller, which rejects the object:

```
PodDisruptionBudget.policy is invalid: spec: Invalid value:
minAvailable and maxUnavailable cannot be both set
```

RedisCluster reconciliation then loops on the failed PDB and follower StatefulSets are never created.

## Fix

Guard each field with `{{- if .pdb.<field> }}` so users who explicitly set only one keep the chart behaviour they asked for, while setting both still renders both (preserving current behaviour for deliberately-configured overrides).

## Test plan

- [x] Hand-verified the rendered YAML shape with each of the three configurations (only `maxUnavailable`, only `minAvailable`, both)
- [ ] `helm lint charts/redis-cluster` (no `helm` available in my local env; left for CI)

### DIFF
--- a/charts/redis-cluster/templates/_helpers.tpl
+++ b/charts/redis-cluster/templates/_helpers.tpl
@@ -33,8 +33,12 @@ tolerations:
 {{- if .pdb.enabled  }}
 pdb:
   enabled: {{ .pdb.enabled }}
+{{- if .pdb.maxUnavailable }}
   maxUnavailable: {{ .pdb.maxUnavailable }}
+{{- end }}
+{{- if .pdb.minAvailable }}
   minAvailable: {{ .pdb.minAvailable }}
+{{- end }}
 {{- end }}
 {{- if .nodeSelector }}
 nodeSelector:


SAY-5:
Good catch, both charts had the same shape. Pushed the same conditional to redis-sentinel.yaml and redis-replication.yaml so a null minAvailable or maxUnavailable no longer ends up in the CR.

Rendered both with pdb.enabled=true in each direction to confirm only the set field shows up.

PR #1751, opened 2026-04-21
fix: use lowercase appendonly.aof to match legacy data

### LINKED ISSUE #1746: Enabling GenerateConfigInInitContainer on existing Cluster will delete all data

redis-operator version: v0.24.0

Yes

**What did you do?**

Setting `GenerateConfigInInitContainer` feature gate from `false` to `true` on an existing redis-operator installation with redis deployed.

Redis to be redeployed with init-container and existing appendaof files loaded.

**What did you see instead?**

Redis will re-initialize without any data, losing all keys.

The Problem is that with the feature gate enabled the `appendfilename` Redis configuration option is set to `Appendonly.aof`. However with the old approach the file name is lowercase (`appendonly.aof`).

https://github.com/OT-CONTAINER-KIT/redis-operator/blob/44479b3d7140188a061e7814a465444b59cbd43c/internal/agent/bootstrap/redis/config.go#L129

This causes Redis on startup to not find the existing files and re-create them from scratch.

We should make sure both approaches use the same file name, however changing this now is a bit more tricky. We might want to make the filename user configurable or implement some automatic conversion.

In any case this should be probably mentioned somewhere in the migration notes.

SAY-5:
## Summary

The `GenerateConfigInInitContainer` path wrote `appendfilename` as `Appendonly.aof`, while the legacy controller-rendered config uses lowercase `appendonly.aof`. Switching the feature gate from `false` to `true` on an existing cluster silently destroyed the AOF: on case-sensitive filesystems redis started up, failed to open the capitalized file, treated the data directory as empty, and re-initialized.

## Fix

Lowercase the filename so both code paths reference the same file and existing AOFs survive the migration.

Fixes #1746.

## Test

- `go build ./internal/agent/bootstrap/...` green.
- `gofmt` clean.

Signed-off-by: SAY-5 <SAY-5@users.noreply.github.com>

### DIFF
--- a/internal/agent/bootstrap/redis/config.go
+++ b/internal/agent/bootstrap/redis/config.go
@@ -126,7 +126,15 @@ func GenerateConfig() error {
 		cfg.Append("save", "300 10")
 		cfg.Append("save", "60 10000")
 		cfg.Append("Appendonly", "yes")
-		cfg.Append("Appendfilename", "\"Appendonly.aof\"")
+		// Use lowercase "appendonly.aof" so the init-container and the
+		// legacy redis-operator controller agree on the AOF file name.
+		// Existing clusters created with the non-init-container path
+		// write appendonly.aof; switching GenerateConfigInInitContainer
+		// from false to true on a running cluster used to point redis at
+		// Appendonly.aof instead, which (on case-sensitive filesystems)
+		// meant redis found no AOF on startup, re-initialized empty, and
+		// silently destroyed the data. See OT-CONTAINER-KIT/redis-operator#1746.
+		cfg.Append("Appendfilename", "\"appendonly.aof\"")
 		cfg.Append("dir", dataDir)
 	} else {
 		fmt.Println("Running without persistence mode")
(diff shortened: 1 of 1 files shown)


internal/agent/bootstrap/redis/config.go
```diff
@@ -126,7 +126,15 @@ func GenerateConfig() error {
...
 		cfg.Append("Appendonly", "yes")
-		cfg.Append("Appendfilename", "\"Appendonly.aof\"")
+		// Use lowercase "appendonly.aof" so the init-container and the
+		// legacy redis-operator controller agree on the AOF file name.
+		// Existing clusters created with the non-init-container path
+		// write appendonly.aof; switching GenerateConfigInInitContainer
+		// from false to true on a running cluster used to point redis at
+		// Appendonly.aof instead, which (on case-sensitive filesystems)
+		// meant redis found no AOF on startup, re-initialized empty, and
+		// silently destroyed the data. See OT-CONTAINER-KIT/redis-operator#1746.
```
shubham-cmyk: Remove these comment, i think we don't need that and too verbose

Issue #1719, opened 2026-03-18
RedisReplication and RedisSentinel cannot select master pod


hardy6413:
**What version of redis operator are you using?**
0.23.0

redis-operator version:

**Does this issue reproduce with the latest release?**
yes

**What operating system and processor architecture are you using (`kubectl version`)?**

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version                                                     
Client Version: v1.34.1
Kustomize Version: v5.7.1
Server Version: v1.33.5-gke.2392000

</pre></details>

**What did you do?**

Sentinel Chart Version: 0.16.12
Redis Replication Chart Version:  0.16.13

I was performing failover mechanism tests of a classic redis sentinel deployment with 3 redis pods and 3 sentinel pods.

Reproduction steps
1. Make sure that the solution is running properly - Redis pods are aware of each other clients can connect to the redis. 1 master and 2 slaves present.
2. Try to kill the pod which currently is a master to see if failover happens.
```bash
while true; do
  kubectl delete pod redis-sentinel-2 redis-2 --force
  sleep 5
done
```
3.  Failover to other redis pod should happen. There should be one Master one Slave and 1 dead pod.
4. Add  the current master pod to the delete loop
```bash
while true; do
  kubectl delete pod redis-sentinel-0 redis-0 --force
  sleep 5
done
```
5. Now only 1 redis pod should be alive and it should be slave. other 2 pods should be dead.
6. Try to release from the loop one of the redis pods
7. Now 2 redis pods should be alive ✅  but they are not aware of each other ❌ 
8. After some time release another pod which was killed. Now 3 redis pods should be alive  ✅  but they are not aware of each other ❌ 
9. When we inspect the roles of the redis pod with this command
```bash
for i in {1..30}; do
  echo "--- try $i ---"
  for p in redis-0 redis-1 redis-2; do
    echo "=== $p ==="
    kubectl -n "$NS" exec "$p" -c redis -- sh -c \
      "REDISCLI_AUTH='$PASS' redis-cli INFO replication | egrep 'role:|master_host:|master_link_status:'"
  done
  sleep 2
done
```

We observe  such output
```text
=== redis-0 ===
role:master
=== redis-1 ===
role:slave
master_host:192.168.5.147
master_link_status:down
=== redis-2 ===
role:master
```
 

10.  Redis pods are not aware of each other. When we inspect the master pods
```bash
[...]

Issue #1711, opened 2026-03-12
RedisReplication controller cannot converge after Sentinel failover when one Redis pod is Pending/unreachable (master service remains stale)


sinner-:
**What version of redis operator are you using?**
- chart_version: `v0.23.0`
- operator_version: `v0.23.0`
- redis image: `quay.io/opstree/redis:v8.4.2`
- sentinel image: `quay.io/opstree/redis-sentinel:v8.4.2`
- exporter image: `quay.io/opstree/redis-exporter:v1.81.0`

**Does this issue reproduce with the latest release?**

Yes, reproducible on `v0.23.0` (chart + operator).

**What operating system and processor architecture are you using (`kubectl version`)?**

1.35.1 x86 Ubuntu 24.04

**What did you do?**

I deployed Redis using separate `RedisReplication` and `RedisSentinel` CRs (3 replicas each), with:
- pod anti-affinity across zones
- local storage (`topolvm-provisioner`)
- app clients pointing to `<name>-master` service (`sina-dev-redis-master`)

Repro steps:
1. Ensure cluster is healthy (`1 master + 2 replicas`) and Sentinel is monitoring.
2. Cordon the node hosting the current Redis master pod.
3. Delete the current master pod.
4. Due to local PV + cordoned node, deleted pod stays `Pending` (expected in this environment).
5. Sentinel elects a new master quickly.
6. Observe app connectivity and operator logs.

```
kind: RedisReplication
metadata:
  name: sina-dev-redis
  namespace: sina-dev
spec:
  affinity:
    podAntiAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
      - labelSelector:
          matchLabels:
            app: sina-dev-redis
        topologyKey: topology.kubernetes.io/zone
  clusterSize: 3
  kubernetesConfig:
    image: quay.io/opstree/redis:v8.4.2
    resources:
      limits:
        cpu: "1"
        memory: 256Mi
      requests:
        cpu: 50m
        memory: 256Mi
    updateStrategy: {}
  pdb:
    enabled: true
    minAvailable: 2
  redisExporter:
    enabled: true
    image: quay.io/opstree/redis-exporter:v1.81.0
    port: 9121
  storage:
    volumeClaimTemplate:
      metadata: {}
      spec:
        accessModes:
        - ReadWriteOnce
        resources:
          requests:
            storage: 5Gi
        storageClassName: topolvm-provisioner
      status: {}
[...]

Issue #1635, opened 2026-01-14
Sentinel PODs hold onto older IPs of replication PODs resulting in accessibility


aneja17:
What version of redis operator are you using?

redis-operator version:
v0.21.2 

What did you do?

Setup:

RedisReplication with clusterSize 3
RedisSentinel with clusterSize 3

Steps to reproduce:

Redis is being used as a cache by one service in our production environment. The service restarts all the replication and sentinel PODs. After the PODs were restarted, sentinel logs show it tries to connect to master replication PODs on old IP address than the new one.

What did you expect to see?

Sentinel configuration updates the new IP addresses of the replication PODs so the connectivity doesn't break and sentinel doesn't hold onto old IPs.

What did you see instead?

Failover stuck and inaccessible redis.

What seems to be the problem:

Sentinel holding onto old IP addresses.

Some logs:
```
MacBook-Pro ~ % kubectl logs arsched-cache-sentinel-1 -n arsched --context infra

Defaulted container "arsched-cache-sentinel" out of: arsched-cache-sentinel, redis-exporter

Sentinel is running without password which is not recommended

ACL_MODE is not true, skipping ACL file modification

Running sentinel without TLS mode

Starting  sentinel service .....

1:X 13 Jan 2026 08:41:44.253 # oO0OoO0OoO0Oo Redis is starting oO0OoO0OoO0Oo

1:X 13 Jan 2026 08:41:44.253 # Redis version=6.2.20, bits=64, commit=00000000, modified=0, pid=1, just started

1:X 13 Jan 2026 08:41:44.253 # Configuration loaded

1:X 13 Jan 2026 08:41:44.253 * monotonic clock: POSIX clock_gettime

1:X 13 Jan 2026 08:41:44.254 * Running mode=sentinel, port=26379.

1:X 13 Jan 2026 08:41:44.256 # Sentinel ID is db71ca9edb61959a47ac6db4f0e78a0604ba5ccd

1:X 13 Jan 2026 08:41:44.256 # +monitor master mymaster 0.0.0.0 6379 quorum 2

1:X 13 Jan 2026 08:41:52.385 # -monitor master mymaster 0.0.0.0 6379

1:X 13 Jan 2026 08:41:52.389 # +monitor master mymaster 10.0.149.114 6379 quorum 2

1:X 13 Jan 2026 08:41:52.389 # +set master mymaster 10.0.149.114 6379 auth-pass ******

[...]

PR #1449, opened 2025-07-21
fix: Recover redisreplication from master pod deletion

dcaputo-harmoni:
**Description**

When the master of a redisreplication is deleted, the slaves continue to point to the IP address of the old master until the pods are restarted.

Fixes [NO ISSUE EXISTS - BUT DESCRIPTION OF ISSUE IS BELOW]

We encountered an issue where the node that a master pod of a redisreplication was running on was cycled, and the master pod was evicted / killed as a result.  The cluster never recovered, because the slaves were all pointing to the old master.

**Type of change**

* Bug fix (non-breaking change which fixes an issue)

**Checklist**

- [ ] Tests have been added/modified and all tests pass.
- X Functionality/bugs have been confirmed to be unchanged or fixed.
- X I have performed a self-review of my own code.
- X Documentation has been updated or added where necessary.

**Additional Context**

The operator was previously only set up to handle a split brain (i.e. multiple master) scenario, so I've extended the logic to handle a orphaned master scenario as well.

PR #1780, opened 2026-05-28
fix(redisreplication): re-point slaves with stale master_host (#1779)

### LINKED ISSUE #1779: RedisReplication+Sentinel: slaves keep stale master_host after master pod IP change — silent SPOF
## Summary

When a `RedisReplication` master pod is recreated with a new IP (node reboot, eviction, pod delete), the operator never re-points the slave pods at the new master. Slaves keep `master_host: <stale_old_ip>`, `master_link_status: down` indefinitely. Sentinel sees the master correctly and `Service/redis-master` endpoint stays healthy, so application writes succeed — but the cluster runs with **zero functioning replicas**. Silent SPOF until something else surfaces it.

We hit this in production: slaves were broken for **~19 hours** before noticed during unrelated work. The operator logged `"Sentinel has incorrect number of slaves, reset needed"` continuously throughout, but the recovery code path is gated to only run when sentinel is disabled.

## Root cause (with code references)

Per `internal/controller/redisreplication/redisreplication_controller.go:397-413`:

```go
} else if len(masterNodes) == 1 && len(slaveNodes) > 0 {
    ...
    if currentRealMaster == \"\" && !instance.EnableSentinel() {
        // CreateMasterSlaveReplication (the only branch that re-issues SLAVEOF)
```

The slave-reconfigure branch is gated by `!instance.EnableSentinel()` — so with sentinel enabled (the recommended production setup), **operator deliberately defers slave reconfiguration to sentinel**. But sentinel only sends `SLAVEOF` to slaves during a quorum-confirmed `+sdown`/`+odown` failover. On a graceful pod-IP swap (no failover, just a new IP for the same master), sentinel updates its own view via the `+switch-master` event… but slaves never receive `REPLICAOF` because no failover happened.

Additionally:
- `CreateMasterSlaveReplication` (`internal/k8sutils/redis.go:767-803`) iterates only over `masterPods` (the role-master-reporting pods, for split-brain handling). It never iterates `slaveNodes` to fix their stale `master_host`.
- The only `SLAVEOF` call site in the entire repo is line 794 inside that function. Verified with `grep -rni \"slaveof\|replicaof\"`.
- The `sentinelResetIfNeed` path (controller.go:300-345) reaches `SentinelReset` at the **sentinel pod** (port 26379) on detecting `masterInfo.Slaves != expectedSlaves`. That only purges sentinel's known-slaves cache — sentinel re-discovers slaves via the master's `INFO replication`, but the new master sees no slaves (because slaves are connecting to dead old IP), so the cycle repeats forever every ~30s.

## Reproduction

1. Healthy 3-pod `RedisReplication` + 3-pod `RedisSentinel` (separate CR or embedded). Confirm `redis-cli -a \$PASS INFO replication` on master shows `connected_slaves: 2`.
2. Delete the master pod: `kubectl delete pod <master-pod-name>`. (Or reboot the node hosting it.)
3. New master pod comes up with a new IP. `Service/redis-master` endpoint is correctly updated by sentinel.
4. Wait a few minutes. Then on each slave pod: `redis-cli -a \$PASS INFO replication | grep master_host` → still the OLD IP.
5. On the new master: `redis-cli -a \$PASS INFO replication | grep connected_slaves` → `0`.
6. Operator logs every ~30s:
   ```
   {\"level\":\"info\",\"msg\":\"Sentinel has incorrect number of slaves, reset needed\",\"expected\":2,\"actual\":0}
   ```
7. State persists indefinitely until manual intervention or a slave pod restart.

## Manual workaround

```bash
MASTER_IP=\$(kubectl exec -n <ns> <sentinel-pod> -- \\
  redis-cli -p 26379 SENTINEL get-master-addr-by-name <mastername> | head -1 | tr -d '\\r')
for slave in <slave-pods>; do
  kubectl exec -n <ns> \$slave -- sh -c 'redis-cli -a \"\$REDIS_PASSWORD\" --no-auth-warning REPLICAOF '\\\"\$MASTER_IP\\\"' 6379'
done
```

After this, `master.connected_slaves` returns to 2 within seconds. `masterauth` is already set by the operator (matches `requirepass`).

## Why this hurts in production

Standard monitoring does NOT catch this:
- `redis_up{role=\"master\"}` is fine (master is fine)
- `redis_master_link_up` on slaves is unreliable — it reports the slave's *attempt* status, can show stale values
- `Service/redis-master` endpoint count is fine (sentinels do their job)
- Application reads/writes work normally (they hit master via Service)

The only direct signal is `master.connected_slaves == 0` for an unexpected period — which most ops teams don't alert on (the OSS chart's `redis-alerts` rule even had a structural bug for years that prevented this firing — separate issue, but means many users won't have caught it).

If the master then dies (planned restart, node fail, OOM), there is no slave to promote → sentinel cannot complete failover → `Service/redis-master` endpoint goes to 0 → apps fail with `EPERM` (kube-proxy REJECT-on-empty-Service). Single-node Redis with the operational illusion of HA.

## Proposed fix

Add a new helper `k8sutils.RepointStaleSlaves(ctx, client, cr, currentMasterPod, slavePods)`:

1. Resolve `currentMasterPod` → IP (using existing `getRedisServerIP`).
2. For each `slavePod`: open a client via `configureRedisReplicationClient` (handles TLS, auth, hostname-mode).
3. `INFO replication` on the slave → parse `master_host`.
4. If `master_host != currentMasterIP` (and slave currently reports `role:slave`), issue `SLAVEOF currentMasterIP 6379` via the existing go-redis client.
5. Log outcome; best-effort (log-and-continue per repo precedent in `configureSentinel`); do not fail the reconcile.

Invocation in `reconcileRedis` (controller.go around line 424), called **before** `configureSentinel`, **unconditional** of `EnableSentinel`:

```go
if realMaster != \"\" && len(slaveNodes) > 0 {
    k8sutils.RepointStaleSlaves(ctx, r.K8sClient, instance, realMaster, slaveNodes)
}
```

### Why unconditional, not gated on \`masterInfo.Slaves != expected\`

The narrower gate would miss cases where a slave is in the connected-slaves count from a sentinel's POV but actually configured against a different/stale IP (we observed this during recovery: sentinel reports 2 slaves, but only 1 is actually replicating). The cheap `INFO replication` per reconcile is acceptable cost; idempotent (no `REPLICAOF` if `master_host` already matches).

### Code style precedent

Mirrors PR #1449 (`fix: Recover redisreplication from master pod deletion`): new helper in `internal/k8sutils/`, called from `reconcileRedis`, structured logging, `log.Error(...)` swallow-and-continue pattern. PR #1705 (`Repair Disconnected Nodes` for RedisCluster) shows the architecturally correct pattern — this proposal ports the same idea to RedisReplication.

## Cross-referenced existing issues (all describing variants of the same root cause)

- #1052 — \"Redis can't recover after node is down with redis-replicas + HA sentinel\" (many +1s through Mar 2026, multiple versions confirmed broken)
- #1314 — \"Badly timed and unnecessary CreateMasterSlaveReplication breaks redis replication\"
- #1635 — \"Sentinel PODs hold onto older IPs of replication PODs\"
- #1711 — \"RedisReplication controller cannot converge after Sentinel failover\"
- #1719 — \"RedisReplication and RedisSentinel cannot select master pod\" (user workaround in comments is literally the missing operator behavior)
- #1179 — \"Follower cannot connect to master because of bad IP\"
- PR #1449 (merged) — incomplete fix, gated by \`!EnableSentinel()\`
- PR #1705 (open) — right pattern, for \`RedisCluster\` only

## Environment

- Operator: `quay.io/opstree/redis-operator:v0.25.0`
- Helm chart: `redis-operator 0.24.0` (CRD `redis.redis.opstreelabs.in/v1beta2`)
- Redis: `quay.io/opstree/redis:v7.4.8`
- Kubernetes: k3s v1.35.2 (3-node bare metal)
- Reliably reproducible.

Happy to provide additional logs, test against a proposed fix, or open the PR.

wowsel:
Fixes #1779.

## Problem

When a `RedisReplication` master pod is recreated with a new IP (graceful pod restart, node reboot, eviction — anything short of a sentinel-detected `+sdown` failover), the operator currently never re-points slaves at the new master. Slaves keep `master_host: <stale_old_ip>`, `master_link_status: down` indefinitely. Master reports `connected_slaves: 0`. Sentinel sees the master correctly via the freshly-issued `SENTINEL MONITOR` and updates the master Service endpoint — so apps think everything is healthy, but the cluster has zero functioning replicas. Silent SPOF.

Investigation summary:
- `SentinelReset` (the only recovery action on detected `Sentinel has incorrect number of slaves`) targets sentinel pods on port 26379 — it purges sentinel's known-slaves cache; it does NOT touch slave data pods.
- The existing slave-reconfigure branch in `reconcileRedis` (`internal/controller/redisreplication/redisreplication_controller.go:401`) is gated by `!instance.EnableSentinel()`, so sentinel-enabled (recommended HA) deployments never take that path.
- `CreateMasterSlaveReplication` iterates only over role:master pods (for split-brain handling); never iterates slaves with stale `master_host`.
- `grep -rni \"slaveof\|replicaof\" internal/ pkg/` confirms exactly one `SLAVEOF` call site in the entire codebase, gated as above.

Production reproducer + 19-hour silent SPOF in #1779.

## Change

Adds a small best-effort helper `k8sutils.RepointStaleSlaves` that, on every reconcile when a real master is known and slaves exist, queries `INFO replication` on each slave and issues `SLAVEOF <current-master-addr> 6379` if the live `master_host` disagrees with the operator-known master.

- **Reuses `configureRedisReplicationClient`** so TLS, `ExistingPasswordSecret`, and the IP-vs-hostname switch are all handled identically to the existing `SLAVEOF` call in `CreateMasterSlaveReplication`.
- **Best-effort:** per-pod failures are logged and skipped, never fail the reconcile (matches the `configureSentinel` precedent).
- **Idempotent:** `SLAVEOF` only issued on actual mismatch — cheap `INFO` otherwise.
- **Unconditional trigger** (not gated on \"Sentinel has incorrect number of slaves\"): that per-sentinel-view signal can lie when a slave is registered with a sentinel but actually replicating from a different/stale IP. The cheap `INFO` per reconcile is acceptable cost.

## Test plan

- [ ] CI passes (this PR adds no new tests; the existing `CreateMasterSlaveReplication` / `GetRedisReplicationRealMaster` are also untested at the unit-test level — happy to add miniredis-based tests if you'd like, just want to confirm the approach first since it would mean introducing a new test infra pattern)
- [ ] Manual repro on 3-node cluster:
  1. Healthy RedisReplication + Sentinel, `connected_slaves: 2`
  2. `kubectl delete pod <master-pod>`
  3. **Before this PR:** slaves stuck on old IP, `connected_slaves: 0` indefinitely
  4. **After this PR:** within one reconcile (~30s) operator logs `\"Stale master_host on slave, issuing SLAVEOF to current master\"` and slaves recover

I will run the manual repro on my prod cluster (k3s v1.35.2, 3 node bare metal) once a build with this fix is available — happy to confirm and provide log evidence.

## Scope (deliberately small)

- Does NOT touch the existing `!EnableSentinel()` recovery branch — that path is a separate question (#1314, #1052)
- Does NOT change `SentinelReset` behavior
- Does NOT modify `CreateMasterSlaveReplication`

The helper is small and additive; risk profile is per-pod idempotent `SLAVEOF` calls when `INFO replication.master_host` mismatches the operator-computed master address. If `realMaster` is empty or no slaves exist, function returns immediately.

Cross-references: #1052, #1179, #1314, #1635, #1711, #1719 all describe variants of this same root cause class. #1449 was a partial fix gated to non-sentinel setups. #1705 ports the right pattern (per-reconcile repair) to RedisCluster — this PR ports the analogous (smaller) pattern to RedisReplication.

### DIFF
--- a/internal/controller/redisreplication/redisreplication_controller.go
+++ b/internal/controller/redisreplication/redisreplication_controller.go
@@ -420,6 +420,14 @@ func (r *Reconciler) reconcileRedis(ctx context.Context, instance *rrvb2.RedisRe
 	monitoring.RedisReplicationReplicasSizeCurrent.WithLabelValues(instance.Namespace, instance.Name).Set(float64(len(masterNodes) + len(slaveNodes)))
 	monitoring.RedisReplicationReplicasSizeDesired.WithLabelValues(instance.Namespace, instance.Name).Set(float64(*instance.Spec.Size))
 
+	// Re-point any slave whose live master_host differs from the current
+	// master. Covers the graceful master-pod-IP-change case where sentinel
+	// updates its own view but never sends SLAVEOF to slaves (no failover
+	// event). Best-effort. See #1779.
+	if realMaster != "" && len(slaveNodes) > 0 {
+		k8sutils.RepointStaleSlaves(ctx, r.K8sClient, instance, realMaster, slaveNodes)
+	}
+
 	if instance.EnableSentinel() {
 		if err := r.configureSentinel(ctx, instance, realMaster); err != nil {
 			log.FromContext(ctx).Error(err, "failed to configure sentinel")
--- a/internal/k8sutils/redis.go
+++ b/internal/k8sutils/redis.go
@@ -802,6 +802,63 @@ func CreateMasterSlaveReplication(ctx context.Context, client kubernetes.Interfa
 	return nil
 }
 
+// RepointStaleSlaves issues SLAVEOF to any slave pod whose live master_host
+// differs from the operator-known current master. Best-effort: failures are
+// logged and skipped, never fail the reconcile. Addresses #1779 — silent
+// SPOF when master pod IP changes (graceful pod recreation, not a sentinel
+// failover) and slaves keep replicating from the dead old IP.
+func RepointStaleSlaves(ctx context.Context, client kubernetes.Interface, cr *rrvb2.RedisReplication, currentMasterPod string, slavePods []string) {
+	if currentMasterPod == "" || len(slavePods) == 0 {
+		return
+	}
+
+	currentMasterInfo := RedisDetails{
+		PodName:   currentMasterPod,
+		Namespace: cr.Namespace,
+	}
+
+	var currentMasterAddr string
+	if cr.Spec.TLS != nil {
+		currentMasterAddr = getRedisReplicationHostname(currentMasterInfo, cr)
+	} else {
+		currentMasterAddr = getRedisServerIP(ctx, client, currentMasterInfo)
+		if currentMasterAddr == "" {
+			log.FromContext(ctx).V(1).Info("RepointStaleSlaves: empty current master IP, skipping", "master", currentMasterPod)
+			return
+		}
+	}
+
+	for _, podName := range slavePods {
+		redisClient := configureRedisReplicationClient(ctx, client, cr, podName)
+		info, err := redisClient.Info(ctx, "Replication").Result()
+		if err != nil {
+			log.FromContext(ctx).Error(err, "RepointStaleSlaves: INFO replication failed", "pod", podName)
+			redisClient.Close()
+			continue
+		}
+
+		var currentHost string
+		for _, line := range strings.Split(info, "\r\n") {
+			if v, ok := strings.CutPrefix(line, "master_host:"); ok {
+				currentHost = strings.TrimSpace(v)
+				break
+			}
+		}
+
+		if currentHost == "" || currentHost == currentMasterAddr {
+			redisClient.Close()
+			continue
+		}
+
+		log.FromContext(ctx).Info("Stale master_host on slave, issuing SLAVEOF to current master",
+			"pod", podName, "stale", currentHost, "current", currentMasterAddr)
+		if err := redisClient.SlaveOf(ctx, currentMasterAddr, "6379").Err(); err != nil {
+			log.FromContext(ctx).Error(err, "RepointStaleSlaves: SLAVEOF failed", "pod", podName, "master", currentMasterAddr)
+		}
+		redisClient.Close()
+	}
+}
+
 func GetRedisReplicationRealMaster(ctx context.Context, client kubernetes.Interface, cr *rrvb2.RedisReplication, masterPods []string) string {
 	for _, podName := range masterPods {
 		redisClient := configureRedisReplicationClient(ctx, client, cr, podName)


wowsel:
While reviewing the fix on our cluster, found the same class of bug also affects the sentinel controller — sentinels stay configured against the dead master IP after a graceful master pod restart, because `RedisSentinelReconciler.SetupWithManager` doesn't watch pods either. Filed as #1781. Keeping this PR scoped to slaves, sentinel fix would be a separate PR.

honghainguyen777:
Hey, any follow-ups on this? I am getting the same problem. Currently, we have to run a CronJob to reset the any stale replicas, but it looks dirty to our deployment :(

diogosilva30:
Reproduced this independently in a DR simulation: RedisReplication (3) + standalone RedisSentinel (3), quorum 2, TLS+ACL enabled, redis-operator v0.26.0. Restarting all data and sentinel pods at once left slaves pointing SLAVEOF at pod IPs that no longer existed, and nothing re-pointed them automatically. Matches #1883. Looks like one of a few pieces needed here; we also hit #1807 and #1888 in the same test.

PR #1798, opened 2026-06-12
feat: add status subresource to Redis standalone

### LINKED ISSUE #638: [Feature] : Add status sub resource
**Describe the solution you'd like**

There is already status sub resource exist for the redis cluster, It would be good if we support same for the redis standalone, replication and senitnel mode. 

redis-operator version: v0.15.1

shubham-cmyk:
**Description**

Add a simple status to the standalone Redis CRD

Fixes #638 

**Type of change**

**Checklist**

**Additional Context**

### DIFF
--- a/internal/controller/redis/redis_controller.go
+++ b/internal/controller/redis/redis_controller.go
@@ -18,6 +18,7 @@ package redis
 
 import (
 	"context"
+	"reflect"
 
 	rvb2 "github.com/OT-CONTAINER-KIT/redis-operator/api/redis/v1beta2"
 	"github.com/OT-CONTAINER-KIT/redis-operator/internal/controller/common"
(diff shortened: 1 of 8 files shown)


internal/controller/redis/redis_controller.go
```diff
@@ -60,17 +64,83 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	if err = k8sutils.AddFinalizer(ctx, instance, RedisFinalizer, r.Client); err != nil {
 		return intctrlutil.RequeueE(ctx, err, "failed to add finalizer")
 	}
+	if instance.Status.State == "" || instance.Status.State == rvb2.RedisFailed {
```
joshsouza, line 67: Your tests are failing due to a race condition/infinite loop here.
If you reach a failed state, on next reconcile you reach this block, which changes you from the failed state to an initializing state and re-enqueues, which will presumably then reach a failed state on the next reconcile. So a failure status is stuck flapping (and your test properly caught that it's not expected behavior).

I believe that if you leave it in the failed state it will continue to be re-enqueued to reprocess as desired and if it is a fixable state, recover, so this change should do the trick.

Thanks for all this! Would really love to see this feature go out.

```suggestion
	if instance.Status.State == "" {
```

PR #1807, opened 2026-06-17
Fix Sentinel bootstrap writing 0.0.0.0 as monitored master

### LINKED ISSUE #1806: Sentinel bootstrap can monitor master at 0.0.0.0, leading to master-down state
## Bug Report

Sentinel bootstrap can write an active `sentinel monitor` entry with `0.0.0.0` as the master address, which seeds invalid Sentinel state and can lead to `master-down` / `0.0.0.0` being reported as the monitored master.

## Impact

This can break Sentinel monitoring/failover state, especially on first boot or after restart before runtime reconciliation repairs Sentinel state.

## Root Cause

The Sentinel bootstrap config generation defaults `IP` to `0.0.0.0` and unconditionally emits:

`sentinel monitor <master-group> 0.0.0.0 6379 <quorum>`

That means invalid state is written directly into `sentinel.conf`.

## Reproduction

1. Start Sentinel without an `IP` env var set.
2. Inspect the generated `sentinel.conf`.
3. Observe an active monitor line using `0.0.0.0` as the master address.
4. After startup/restart, Sentinel may report the master as `0.0.0.0` and mark it down.

## Expected

Bootstrap should not emit `sentinel monitor` unless it has a valid master address.

## Actual

Bootstrap can write `0.0.0.0` as the monitored master.

## Proposed Fix

- Do not default Sentinel bootstrap `IP` to `0.0.0.0`
- Skip `sentinel monitor` when `IP` is empty or `0.0.0.0`
- Skip `sentinel announce-ip` for the same invalid cases

## Additional Note

There is also an operator-side hardening opportunity for embedded Sentinel in `RedisReplication`: pass `MASTER_GROUP_NAME`, `PORT`, and Sentinel tuning env vars consistently into the Sentinel container bootstrap. That is useful, but the source bug is the invalid bootstrap default described above.

barrowkwan:
## Summary

Fixes OT-CONTAINER-KIT/redis-operator#1806

## Root Cause

Sentinel bootstrap defaulted `IP` to `0.0.0.0` and unconditionally wrote:

`sentinel monitor <group> 0.0.0.0 6379 <quorum>`

That causes invalid Sentinel state on first boot or restart before runtime reconciliation repairs it.

## Changes

- stop defaulting bootstrap `IP` to `0.0.0.0`
- only emit `sentinel monitor` when `IP` is a real address
- only emit `sentinel announce-ip` when `IP` is valid
- add bootstrap tests for:
  - unset `IP`
  - `IP=0.0.0.0`
  - valid `IP`
- harden embedded Sentinel bootstrap env wiring in the replication controller by passing:
  - `MASTER_GROUP_NAME`
  - `PORT`
  - Sentinel timing/tuning env vars
  - hostname settings env vars

## Why

The source bug is in Sentinel bootstrap. Operator reconciliation can repair state later, but it should not need to undo invalid bootstrap config.

## Validation

- `go test ./internal/agent/bootstrap/sentinel`
- `go test ./internal/controller/redisreplication ./internal/controller/redissentinel`

## Scope

The real source fix belongs in Sentinel bootstrap/image behavior. The operator-side embedded Sentinel env change is additional hardening.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,6 +47,7 @@ CRD_REF_DOCS = $(LOCALBIN)/crd-ref-docs-$(CRD_REF_DOCS_VERSION)
 KUSTOMIZE_VERSION ?= v5.6.0
 CONTROLLER_TOOLS_VERSION ?= v0.17.2
 ENVTEST_VERSION ?= release-0.17
+ENVTEST_USE_DEPRECATED_GCS ?= false
 GOLANGCI_LINT_VERSION ?= v2.4.0
 KUTTL_VERSION ?= 0.15.0
 KIND_VERSION ?= v0.24.0
@@ -71,7 +72,7 @@ all: manager
 # Run tests
 .PHONY: test
 test: generate fmt vet manifests
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go test ./... -coverprofile cover.out
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --use-deprecated-gcs=$(ENVTEST_USE_DEPRECATED_GCS) -p path)" go test ./... -coverprofile cover.out
 
 # Build manager binary
 .PHONY: manager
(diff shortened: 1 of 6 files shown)


diogosilva30:
Hit this too in a simultaneous full-pod-restart DR test (RedisReplication + RedisSentinel, v0.26.0, quorum 2, TLS). A restarted sentinel pod came back monitoring mymaster 0.0.0.0 6379 and never rejoined quorum until we manually repaired it with SENTINEL REMOVE/MONITOR/SET/RESET. Same failure class as #1883.

PR #1827, opened 2026-06-30
feat(cluster): rejoin isolated nodes, reattach empty masters, forget stale IDs

arajni3:
## Summary

Recovers RedisCluster members that return under a **new node ID** (lost `nodes.conf` on pure-cache / ephemeral conf), without relying on the node-count creation path.

Three complementary repairs (all data-safe where they act):

1. **`ForgetStaleNodes`** — forget fail/noaddr orphan IDs that are not the live pod and do not own slots
2. **`RejoinIsolatedNodes`** — per-pod `CLUSTER INFO`; `CLUSTER MEET` when `cluster_known_nodes <= 1` (invisible to gossip-based repair when leader-0 dropped the pod)
3. **`ReattachMisplacedReplicas`** — demote empty (0-slot) masters onto the shard's **live slot owner** via `CLUSTER REPLICATE` on the empty node only (empty syncs from owner, never reverse)

Reconcile order (repair section): disconnected / stale-replication repair → rejoin → **reattach** → empty-master rebalance. Reattach also runs when `CheckRedisNodeCount > desired` (stale orphans after MEET), so it is not stranded in the creation branch.

**Explicit non-goals:** missed promotion where slots are owned only by a dead orphan (needs forced `CLUSTER FAILOVER`); reattach skips that case.

## Tests
- Unit tests for forget / rejoin / reattach (including Outcome-B skip and idempotency)
- Chainsaw (kind CI):
  - `redis-cluster-rejoin-after-delete` (follower)
  - `redis-cluster-rejoin-leader-after-delete` (ex-leader → working replica)
  - `redis-cluster-forget-stale-node`

Physical multi-operator comparison (baselines vs patched) was used to design discriminating injections; not included in this PR (site-specific harness). Happy to summarize methodology in review if useful.

## Metrics
- `rediscluster_rejoin_isolated_attempt`
- `rediscluster_reattach_replica_attempt`
- (forget metric if already present on branch)

## Checklist
- [x] No site-specific cluster/registry defaults in tree
- [x] Chainsaw only for e2e in this PR


arajni3:
Pushed a comment-only clarification on repair ordering: **rejoin is a prerequisite** for both reattach and forget; **reattach and forget are independent** of each other (reattach-before-forget is preference / roles-then-prune, not a hard unlock; forget-before is equally safe and sometimes reaches `nc==desired` sooner).

yelchuridinesh:
Hi @arajni3 any particular reason for this PR to be closed ? I observed this as a gap in the RedisCluster CRD. I was particularly looking for this feature in the operator. Any plans to merge this in main branch and release a new version of Redis Operator ?

PR #1843, opened 2026-07-22
fix: skip unreachable RedisReplication pods during role discovery

### LINKED ISSUE #1842: RedisReplication: node crash leaves pod Running/Ready (stale) — GetRedisNodesByRole fail-fast blocks role label and status updates
### Description

When a **physical/worker node crashes**, Redis pods on that node commonly stay in the API as `Running` with a **stale `Ready=True`** for minutes (kubelet is dead and cannot refresh status; eviction via `NoExecute` often waits ~5 minutes by default). During that window Redis is unreachable, but the operator still treats the pod as probeable.

`GetRedisNodesByRole` then **fail-fasts on the first Redis probe error**, which aborts the whole RedisReplication reconcile loop. Healthy replicas never get role discovery completed, so:

- `status.masterNode` stays stale
- `redis-role` labels on healthy pods are not updated (`reconcileStatus` / `UpdateRedisRoleLabel` never runs)
- replication repair / sentinel reconfiguration may not run

Node hardware failure is a normal production event; the operator should keep reconciling reachable pods instead of blocking the entire loop on one unreachable instance.

### Environment

- Resource: `RedisReplication` (with or without Sentinel — both paths call `GetRedisNodesByRole`)
- Scenario: one worker node down; pod on that node still shows `Running` + stale Ready / PodIP; other replicas healthy
- Also reproducible without killing a node: block Redis only (e.g. iptables drop 6379) while Pod stays Ready
- Operator: current mainline after #1602 / #1712 / #1720 — partial mitigations exist, but this gap remains

### Why #1602 / #1712 are not enough

| Fix | What it does | Gap for node crash |
|-----|--------------|--------------------|
| #1602 `UpdateRedisRoleLabel` | Skip pod on Redis connect error | Never reached if earlier `GetRedisNodesByRole` fails |
| #1712 `IsRedisPodProbeable` | Skip NotReady / no IP / not Running / NotFound | **Does not help** when status is stale `Running`+`Ready=True` after node death |
| #1720 realMaster fallbacks | Help after masters are discovered | Never runs if role discovery fail-fasts first |

Current behavior for a Ready pod whose Redis probe fails (explicitly covered by tests):

```go
// internal/k8sutils/redis.go — getRedisNodesByRole
if !IsRedisPodProbeable(pod) {
    continue
}
podRole, err := probeRole(ctx, pod)
if err != nil {
    return nil, err  // fail-fast: still blocks reconcile
}
```

`TestGetRedisNodesByRoleFailsWhenReadyPodProbeFails` asserts this error path.

Controller reconcilers are sequential; error aborts before status/labels:

```go
// internal/controller/redisreplication/redisreplication_controller.go
reconcilers := []reconciler{
    {typ: "finalizer", ...},
    {typ: "resources", ...},
    {typ: "redis", rec: r.reconcileRedis},   // GetRedisNodesByRole fails here
    {typ: "status", rec: r.reconcileStatus}, // never reached → labels not updated
}
```

`reconcileStatus` also calls `GetRedisNodesByRole` **before** `UpdateRedisRoleLabel`, so even a decoupled status step would still fail on the same probe error unless discovery skips unreachable pods.

### Inconsistency with other code paths

| Code | Unreachable pod behavior |
|------|--------------------------|
| `GetRedisNodesByRole` | **Return error, stop** |
| `GetRedisReplicationRealMaster` / `checkAttachedSlave` | Log error, return -1, **continue** |
| `Healer.UpdateRedisRoleLabel` | Log error, **skip pod**, continue |
| `configureSentinel` (per pod) | Log error, **continue** |

### Impact

- During node-down window (often several minutes), master Service selection via `redis-role` labels can stay wrong
- `status.masterNode` / metrics stay stale
- Error-driven requeue until the bad pod becomes NotReady, loses IP, or is evicted/replaced
- Common HA event (node crash) unnecessarily degrades operator convergence

### Steps to reproduce

1. Deploy `RedisReplication` with 3 replicas (Sentinel optional).
2. Identify current master (e.g. `...-0`).
3. Simulate node failure for that pod’s node (or: drop traffic to Redis on that pod while Pod Status stays `Running`/`Ready`).
4. Confirm API still shows the dead pod as `Running` (often still Ready) with PodIP present.
5. Watch operator logs: `Failed to Get the role Info` for that pod → reconcile requeues with error.
6. Observe healthy pods: `redis-role` labels / `status.masterNode` do not converge until the stale pod is skipped or removed.

### Expected behavior

- Skip unreachable pods during role discovery (log + continue), same tolerance as `UpdateRedisRoleLabel`.
- Continue `reconcileStatus` so healthy pods get correct `redis-role` labels and `status.masterNode` can update from reachable topology.
- Optionally return error only if **all** probeable pods fail (or none can be probed).

### Proposed fix

1. In `getRedisNodesByRole` / `GetRedisNodesByRole`: on `probeRole` / `checkRedisServerRole` error, **log and continue** (do not fail-fast). Optionally error only if every probeable pod failed.
2. (Optional) Ensure `reconcileStatus` can still update labels even when replication repair is skipped due to incomplete topology.
3. (Optional) Align `CreateMasterSlaveReplication` with skip-on-error for unreachable pods.

### Related

- #1602 — Prevent failover label update failure on unreachable pods (label path only)
- #1712 — Skip unprobeable pods (K8s Ready/IP gate; not Redis reachability)
- #1720 — prevent crash loop when all replication pods restart with empty realMaster
- #1449 — Recover redisreplication from master pod deletion without sentinel

Happy to open a PR if maintainers agree with the approach.

leeworker:
Fixes #1842

## Summary

When a worker node crashes, Redis pods can remain `Running`/`Ready` with stale Kubernetes status while Redis is unreachable. `GetRedisNodesByRole` previously fail-fast on the first probe error, which blocked the entire RedisReplication reconcile loop and prevented healthy pods from updating `status.masterNode` and `redis-role` labels.

This change:

- Skips per-pod Redis probe failures in `GetRedisNodesByRole` (log + continue) instead of aborting reconcile
- Removes stale `redis-role` labels from pods that are probeable in Kubernetes but unreachable over Redis, so master Services do not keep routing to dead pods during the stale-Ready window

## Test plan

- [x] `go test ./internal/k8sutils/... -run 'GetRedisNodesByRole|IsRedisPodProbeable'`
- [x] `go test ./internal/controller/common/redis/... -run UpdateRedisRoleLabel`
- [x] `go test ./internal/controller/redisreplication/... -run 'TestReconcile'`

Made with [Cursor](https://cursor.com)

| Situation | How the controller decides |
|---|---|
| All pods observed, exactly one says master | Trust it. Nothing else could be the master. |
| Partial view (some pods unreachable), one says master, **it has replicas attached** | Trust it. A freshly booted standalone never has replicas, so replicas prove it's the real one. |
| Partial view, lone master, no replicas, **Sentinel enabled** | Ask Sentinel. Sentinel promotes masters on its own, so the recorded master might already have been replaced, and a newly promoted master might legitimately have no replicas (e.g. 2-pod setup, old master died). |
| Partial view, lone master, no replicas, **no Sentinel** | Fall back to `Status.MasterNode`. Nothing else can promote behind the controller's back, so if the recorded master says "I'm master," it still is. |
| Several pods say master | First check which has replicas attached. If none does and Sentinel is on, use Sentinel's pick — so a master Sentinel promoted doesn't get demoted just because `Status.MasterNode` is stale. |
| All pods observed, **none** says master | Conclusive: there is no master. Clear `Status.MasterNode`. |
| Partial view, none of the observed pods says master | Inconclusive: the master may be an unobserved pod. Keep the previous `Status.MasterNode`. |

### DIFF
--- a/internal/controller/common/redis/heal.go
+++ b/internal/controller/common/redis/heal.go
@@ -78,6 +78,16 @@ func (h *healer) UpdateRedisRoleLabel(ctx context.Context, ns string, labels map
 		isMaster, err := h.redis.Connect(connInfo).IsMaster(ctx)
 		if err != nil {
 			log.FromContext(ctx).Error(err, "failed to check redis role, skipping pod", "pod", pod.Name)
+			if oldRole := pod.Labels[common.RedisRoleLabelKey]; oldRole != "" {
+				patch := []byte(fmt.Sprintf(`[{"op": "remove", "path": "/metadata/labels/%s"}]`, common.RedisRoleLabelKey))
+				if rErr := retry.RetryOnConflict(retry.DefaultRetry, patchFunc(pod.Name, patch)); rErr != nil {
+					return fmt.Errorf("failed to remove stale pod role label: %w", rErr)
+				}
+				log.FromContext(ctx).Info("removed stale pod role label after probe failure",
+					"pod", pod.Name,
+					"oldRole", oldRole,
+				)
+			}
 			continue
 		}
 		role := common.RedisRoleLabelSlave
--- a/internal/controller/common/redis/heal.go
+++ b/internal/controller/common/redis/heal.go
@@ -39,11 +39,13 @@ type Healer interface {
 	// label to match the pod's real role.
 	//
 	// If a pod can't be reached, its label is normally left as is. The one exception is a pod
-	// labeled as master that is unreachable while another pod already answers as master and has
-	// replicas attached. In that case the old master label is removed, because keeping it could
-	// point the master service at a dead pod. If that removal fails, an error is returned so the
-	// caller retries.
-	UpdateRedisRoleLabel(ctx context.Context, ns string, labels map[string]string, secret *commonapi.ExistingPasswordSecret, tlsConfig *commonapi.TLSConfig) error
+	// labeled as master that is unreachable while another pod already answers as master and
+	// either has replicas attached or is masterPod, the pod the caller has positively identified
+	// as the master (through Sentinel, say, after a failover that left it no replica to attach).
+	// In that case the old master label is removed, because keeping it could point the master
+	// service at a dead pod. If that removal fails, an error is returned so the caller retries.
+	// Pass an empty masterPod when no master has been identified.
+	UpdateRedisRoleLabel(ctx context.Context, ns string, labels map[string]string, secret *commonapi.ExistingPasswordSecret, tlsConfig *commonapi.TLSConfig, masterPod string) error
 }
 
 type healer struct {
@@ -58,7 +60,7 @@ func NewHealer(clientset kubernetes.Interface) Healer {
 	}
 }
 
-func (h *healer) UpdateRedisRoleLabel(ctx context.Context, ns string, labels map[string]string, secret *commonapi.ExistingPasswordSecret, tlsConfig *commonapi.TLSConfig) error {
+func (h *healer) UpdateRedisRoleLabel(ctx context.Context, ns string, labels map[string]string, secret *commonapi.ExistingPasswordSecret, tlsConfig *commonapi.TLSConfig, masterPod string) error {
 	selector := make([]string, 0, len(labels))
 	for key, value := range labels {
 		selector = append(selector, fmt.Sprintf("%s=%s", key, value))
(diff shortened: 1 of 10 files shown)


leeworker:
Hi maintainers,

Could you please approve the workflows when you have a chance?

This PR fixes #1842: when a worker node crashes, Redis pods can remain `Running`/`Ready` with stale status while Redis is unreachable. `GetRedisNodesByRole` previously fail-fast on the first probe error and blocked the whole RedisReplication reconcile loop, so healthy pods could not update `status.masterNode` or `redis-role` labels.

Thanks!

shubham-cmyk:
@copilot review the PR again

shubham-cmyk:
@copilot i have resolved the previous comments can you review the PR again.

shubham-cmyk (approved):

shubham-cmyk:
@copilot resolve all the conversation that have been fixed.

PR #1856, opened 2026-08-07
fix(cluster): wait for master demotion in preStop hook

### LINKED ISSUE #1855: RedisCluster preStop hook never waits for failover to complete
### Environment
- Operator: `redis-operator` **v0.25.0**
- CR: `RedisCluster` (`redis.redis.opstreelabs.in/v1beta2`), `clusterSize: 3` (3 leader + 3 follower), `clusterVersion: v8`
- Redis image: `quay.io/opstree/redis:v8.6.1`
- No TLS, password auth enabled
- Kubernetes: EKS 1.34

## Summary

The `preStop` hook the operator attaches to RedisCluster pods promotes a replica with `redis-cli … CLUSTER FAILOVER` and then **returns immediately**. `CLUSTER FAILOVER` is asynchronous — it replies `+OK` as soon as the replica *accepts* the handshake, not when the replica has actually taken over the slots. Because the hook doesn't wait, the kubelet proceeds to `SIGTERM` the old master while the coordinated failover may still be in flight. If the old master dies mid-handshake, the coordinated failover is aborted and the cluster falls back to **failure-detection** failover (it waits `cluster-node-timeout` before a replica is elected), producing a window in which that shard's slots are unavailable and clients see errors/timeouts.

Net effect: a routine rolling restart (e.g. changing `spec.kubernetesConfig.resources`) causes client-visible disruption on every **master** pod restart, even though the hook was clearly intended to make the handoff graceful.

## The generated preStop hook (copied verbatim from a running leader pod)

```
kubectl -n <ns> get pod <name>-leader-0 \
  -o jsonpath='{.spec.containers[0].lifecycle.preStop.exec.command}'
```

```sh
#!/bin/sh
ROLE=$(redis-cli -h $(hostname) -p ${REDIS_PORT} -a "${REDIS_PASSWORD}" info replication | awk -F: '/role:master/ {print "master"}')

if [ "$ROLE" = "master" ]; then
    BEST_SLAVE=$(redis-cli -h $(hostname) -p ${REDIS_PORT} -a "${REDIS_PASSWORD}" info replication | awk -F: '
        BEGIN { maxOffset = -1; bestSlave = "" }
        /slave[0-9]+:ip/ {
            split($2, a, ",");
            split(a[1], ip_arr, "=");
            split(a[4], offset_arr, "=");
            ip = ip_arr[2];
            offset = offset_arr[2] + 0;
            if (offset > maxOffset) { maxOffset = offset; bestSlave = ip }
        }
        END { print bestSlave }
    ')

    if [ -n "$BEST_SLAVE" ]; then
        redis-cli -h "$BEST_SLAVE" -p ${REDIS_PORT} -a "${REDIS_PASSWORD}" cluster failover
    fi
fi
```

The script issues one `cluster failover` and exits. There is **no** poll/wait for the failover to finish, and the result of the command is not even checked.

## Why this is incorrect

From the [Redis `CLUSTER FAILOVER` docs](https://redis.io/docs/latest/commands/cluster-failover/#implementation-details-and-notes): the command starts a **coordinated** failover — the replica asks the master to pause clients, syncs to the master's replication offset, wins an election, and only then takes over the slots; the old master afterwards unblocks its clients and starts returning redirects. The command returns **before** any of that has happened. So returning from the hook immediately after issuing it does not guarantee — and usually does not achieve — that the failover is complete before the pod is terminated.

Source (as of the versions inspected): this hook is generated by the operator's StatefulSet builder — `generateClusterPreStop` / `GeneratePreStopCommand` in `pkg/k8sutils/statefulset.go`, added in [PR #1264 "Add preStop hook for Redis Cluster failover"](https://github.com/OT-CONTAINER-KIT/redis-operator/pull/1264) (released in v0.20.0).

## Reproduction

1. Deploy the operator and a `RedisCluster` (3 leaders + 3 followers), password auth on.
2. From a pod that can reach the cluster, run a constant write load and count failures (portable, uses only `redis-cli`):

   ```sh
   PW=<password>
   SEED=<name>-leader.<ns>.svc.cluster.local
   total=0; errors=0
   while true; do
     total=$((total+1))
     redis-cli -c -h "$SEED" -p 6379 -a "$PW" -t 3 set "k:$((RANDOM%5000))" v >/dev/null 2>&1 || errors=$((errors+1))
     if [ $((total%200)) -eq 0 ]; then echo "$(date -u +%H:%M:%S) total=$total errors=$errors"; fi
   done
   ```
   (A cluster-aware client — e.g. go-redis `NewClusterClient`, or `redis-cli -c` as above — makes the impact visible via redirects/timeouts.)
3. Trigger a rolling restart:
   ```sh
   kubectl -n <ns> patch rediscluster <name> --type merge \
     -p '{"spec":{"kubernetesConfig":{"resources":{"requests":{"cpu":"260m","memory":"200Mi"}}}}}'
   ```
4. Watch the error counter: it increments in bursts aligned to each **master** pod restart. Slave-only restarts are quiet (the hook is a no-op for replicas), which localizes the problem to the master-termination path.

## Expected behavior

After issuing `CLUSTER FAILOVER`, the hook should **block until the failover has completed** (i.e., the local node has been demoted to a replica) — or until a timeout safely below `terminationGracePeriodSeconds` — before returning and allowing the pod to be terminated.

## Proposed fix

Wait until the local node reports `role:slave` (definitive "failover complete" signal), bounded by a timeout:

```sh
#!/bin/sh
ROLE=$(redis-cli -h "$(hostname)" -p ${REDIS_PORT} -a "${REDIS_PASSWORD}" info replication | awk -F: '/role:master/{print "master"}')
if [ "$ROLE" = "master" ]; then
    BEST_SLAVE=$(...unchanged: pick highest-offset replica...)
    if [ -n "$BEST_SLAVE" ]; then
        redis-cli -h "$BEST_SLAVE" -p ${REDIS_PORT} -a "${REDIS_PASSWORD}" cluster failover
        # wait until THIS node is demoted (failover complete), bounded < terminationGracePeriodSeconds
        i=0
        while [ $i -lt 40 ]; do
            r=$(redis-cli -h "$(hostname)" -p ${REDIS_PORT} -a "${REDIS_PASSWORD}" info replication | awk -F: '/role:/{print $2}' | tr -d '\r')
            [ "$r" = "slave" ] && break
            sleep 1; i=$((i+1))
        done
    fi
fi
```

Additional hardening worth considering:
- Ensure `terminationGracePeriodSeconds` comfortably exceeds the wait budget (e.g. default it to ≥60s for cluster leaders).
- Check the `CLUSTER FAILOVER` result and optionally retry / fall back to `CLUSTER FAILOVER FORCE` if the replica declines.
- Maybe even consider an option to allow override default prestop hook that is set by operator.

We validated the waiting version by patching the generated StatefulSet directly (operator scaled to 0) — it demotes cleanly before terminating and removed the client errors in our tests.

evheniyt-pillar:
**Description**

The RedisCluster `preStop` hook issues `CLUSTER FAILOVER` to the best slave but returns immediately. Redis failover is asynchronous — the command starts a coordinated failover and returns before it completes — so the kubelet can `SIGKILL` the old master mid-handoff. That aborts the coordinated failover and falls back to slower failure detection, causing client-visible disruption (timeouts/errors) on every master pod restart during rolling updates, even though the graceful-handoff mechanism was designed to prevent exactly that.

This change makes the cluster hook wait for the handoff to complete before returning, mirroring the existing `generateReplicationPreStop` behavior:

- After triggering the failover, `generateClusterPreStop` polls the local node's `info replication` and breaks once it reports `role:slave`, so the hook only returns once demotion is confirmed.
- The wait is bounded by `WaitSeconds`, derived from the role's `terminationGracePeriodSeconds` (grace − 10s headroom, clamped to ≥1), so the hook always returns before `SIGKILL`.
- `PreStopWaitSeconds` is wired into the cluster container params for both leader and follower.
- Renamed the shared helper `replicationPreStopWaitSeconds` → `preStopWaitSeconds`, since it now bounds both the cluster and replication hooks.

Fixes #1855

**Type of change**

**Checklist**

**Additional Context**

Follows the same pattern already used for the Sentinel-driven replication `preStop` hook, so cluster and replication now share the same bounded demotion-wait logic. New/updated unit tests: `TestGenerateClusterPreStopContent` asserts the demotion-wait loop is present, `Test_generateRedisClusterContainerParams` asserts `PreStopWaitSeconds` wiring, and `TestPreStopWaitSeconds` covers the (renamed) shared helper. Verified with `go build ./...`, `go vet ./internal/k8sutils/`, and `go test ./internal/k8sutils/`.


evheniyt-pillar:
Tested it on a real cluster - looks good:
```yaml
        exec:
          command:
          - sh
          - -c
          - |-
            #!/bin/sh
            ROLE=$(redis-cli -h $(hostname) -p ${REDIS_PORT}  -a "${REDIS_PASSWORD}"  info replication | awk -F: '/role:master/ {print "master"}')

            if [ "$ROLE" = "master" ]; then
                BEST_SLAVE=$(redis-cli -h $(hostname) -p ${REDIS_PORT}  -a "${REDIS_PASSWORD}"  info replication | awk -F: '
                    BEGIN { maxOffset = -1; bestSlave = "" }
                    /slave[0-9]+:ip/ {
                        split($2, a, ",");
                        split(a[1], ip_arr, "=");
                        split(a[4], offset_arr, "=");
                        ip = ip_arr[2];
                        offset = offset_arr[2] + 0;
                        if (offset > maxOffset) {
                            maxOffset = offset;
                            bestSlave = ip;
                        }
                    }
                    END { print bestSlave }
                ')

                if [ -n "$BEST_SLAVE" ]; then
                    redis-cli -h "$BEST_SLAVE" -p ${REDIS_PORT}  -a "${REDIS_PASSWORD}"  cluster failover

                    for i in $(seq 1 20); do
                        NEW_ROLE=$(redis-cli -h $(hostname) -p ${REDIS_PORT}  -a "${REDIS_PASSWORD}"  info replication | awk -F: '/role:slave/ {print "slave"}')
                        if [ "$NEW_ROLE" = "slave" ]; then
                            break
                        fi
                        sleep 1
                    done
                fi
            fi
```

shubham-cmyk:
@evheniyt-pillar Thanks a lot for the PR, can you resolve the conflict and we have to add specific e2e test in chainsaw

shubham-cmyk (dismissed):

evheniyt-pillar:
hi @shubham-cmyk I see that e2e tests failed, but it looks like they are not related to these changes, can you take a look?

shubham-cmyk (approved):

PR #1859, opened 2026-08-13
fix: prevent non-resource RBAC rules from breaking admin aggregation

### LINKED ISSUE #1858: ClusterRole with aggregate-to-admin: "true" + nonResourceURLs breaks RBAC on other controllers cluster-wide

redis-operator: v0.25.0 (latest release)
OpenShift 4.21 (Kubernetes v1.34.8)

**Summary**

The redis-operator installs a `ClusterRole` that combines the `rbac.authorization.k8s.io/aggregate-to-admin: "true"` aggregation label with a rule containing `nonResourceURLs`. This combination is invalid in Kubernetes and causes **any** controller that aggregates the `admin` ClusterRole into a namespaced `Role` to fail its reconcile loop — breaking RBAC well outside the scope of redis-operator itself.

**Root cause**

Kubernetes' built-in ClusterRole aggregation mechanism collects rules from any `ClusterRole` labeled `rbac.authorization.k8s.io/aggregate-to-admin: "true"` (or `-edit`, `-view`) into the corresponding default `admin`/`edit`/`view` ClusterRole. This is by design and expected.

However, `nonResourceURLs` is only valid on rules bound via a `ClusterRoleBinding` — it can never legally appear in a namespaced `Role`. Any operator/controller that copies aggregated `admin` rules into a namespace-scoped `Role` will therefore try to create a `Role` containing a `nonResourceURLs` rule, which the Kubernetes API server rejects outright:

```
Role.rbac.authorization.k8s.io "<name>" is invalid: rules[N].nonResourceURLs: Invalid value: ["*"]: namespaced rules cannot apply to non-resource URLs
```

We hit this via OpenShift GitOps (ArgoCD), whose application-controller Role reconcile started failing immediately after installing redis-operator's ClusterRole — with no direct relationship between the two components. This makes the failure very hard to trace back to its source.

**Offending rule**

```yaml
- nonResourceURLs:
    - '*'
  verbs:
    - get
```

on a `ClusterRole` also labeled:

```yaml
metadata:
  labels:
    rbac.authorization.k8s.io/aggregate-to-admin: "true"
```

**Impact**

Any cluster component that aggregates the `admin` ClusterRole into namespaced `Role` objects (ArgoCD/OpenShift GitOps application-controller in our case) fails to reconcile after this ClusterRole is applied, since the resulting `Role` object is rejected by the API server. This can silently break RBAC provisioning for unrelated workloads across the cluster.

**Suggested fix**

Remove the `nonResourceURLs` rule from the ClusterRole that carries the `aggregate-to-admin` label, or split it into a separate ClusterRole without any aggregation label, bound explicitly via a dedicated `ClusterRoleBinding` if non-resource URL access is actually required.

LPX-E5BD8:
fixes #1858

aggregate-to-admin ClusterRole has nonResourceURLs: ["*"].
argo copies it into a Role, apiserver rejects, other rbac breaks.

dropped nonResourceURLs. helm test for the combo.


LPX-E5BD8:
Still ready for review. This keeps `nonResourceURLs` rules out of the aggregated admin ClusterRole so namespaced Role copies stay valid. DCO and GitGuardian are green.

LPX-E5BD8:
Rebased onto current main so this isn't sitting behind anymore. Same change: keep `nonResourceURLs` out of the aggregated admin ClusterRole so namespaced Role copies stay valid.

PR #1861, opened 2026-08-16
docs: select only the metrics service in ServiceMonitor examples

### LINKED ISSUE #1833: Service monitor documentation proposals
**Which document needs to be updated?**

https://redis-operator.opstree.dev/docs/monitoring/#servicemonitor

**Expected changes**

People tend to copy the config from documentation, so it would be beneficial to do some small changes:

1.   imagePullPolicy: Always --> IfNotPresent (which is more default)
2. Remove the labels that are not required:
```yml
labels:
    redis-operator: "true"
    env: production
```

3. An actual bug: 
```yml
spec:
  selector:
    matchLabels:
      redis_setup_type: cluster
```
This setup creates double the metrics (at least for replication and sentinel)

This seems to be correct:
```yml
spec:
  selector:
    matchLabels:
      redis_setup_type: sentinel
      app.kubernetes.io/component: metrics
```

redis-operator version:

0.25.0

**Additional context**

vshulcz:
**Description**

Fixes #1833.

The ServiceMonitor examples select services by `redis_setup_type` alone. When the exporter is enabled, `CreateOrUpdateMetricsService` creates a dedicated `<name>-metrics` service that carries the same setup-type labels plus `app.kubernetes.io/component: metrics`. For standalone, replication and sentinel the regular service is created with `epp`, so it exposes the exporter port too, and both services end up scraped — the duplicate metrics reported in the issue. Cluster services are created with `disableMetrics`, so they are not affected, but the extra label is correct there as well.

The examples now match on `app.kubernetes.io/component: metrics` in addition to `redis_setup_type`, and a short paragraph explains why. The `redis-operator: "true"` / `env: production` labels on the ServiceMonitor metadata are dropped as requested in the issue — nothing selects on them.

I did not change `imagePullPolicy: Always` to `IfNotPresent` (point 1 of the issue). Every example in `example/v1beta2/` uses `Always` for `redisExporter`, so changing only the docs would leave the two inconsistent. Happy to do it across the docs and the examples in a follow-up if you want that default flipped.

**Type of change**

### DIFF
--- a/docs/content/en/docs/Monitoring/_index.md
+++ b/docs/content/en/docs/Monitoring/_index.md
@@ -54,6 +54,8 @@ serviceMonitor:
 
 For kubectl related configuration, we may have to create `ServiceMonitor` definition in a YAML manifest and apply it using `kubectl` command.
 
+When the exporter is enabled, the operator creates a dedicated `<name>-metrics` service labelled with `app.kubernetes.io/component: metrics`. For standalone, replication and sentinel setups the regular Redis service exposes the exporter port as well, so a selector that only matches `redis_setup_type` selects both services and Prometheus scrapes the exporter twice. Match on `app.kubernetes.io/component: metrics` as well to select only the metrics service.
+
 ServiceMonitor for Redis cluster setup:
 
 ```yaml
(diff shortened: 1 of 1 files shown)


shubham-cmyk (approved):
LGTM!

PR #1862, opened 2026-08-16
docs: document OPERATOR_STS_PVC_TEMPLATE_NAME

### LINKED ISSUE #1365: Add Documentation Regarding Legacy PVC Naming
**Which document needs to be updated?**
The Upgrading docs and/or the example values.yaml file for the Redis Operator

**Expected changes**
Add documentation regarding the `OPERATOR_STS_PVC_TEMPLATE_NAME` environment variable

redis-operator version: v0.19.1

**Additional context**
We upgraded from a very old version of the redis operator (0.14.0) to v0.19.1. 
It caused us some production issues when pods were deleted due to the change in the VolumeClaimTemplate naming.
We had to go over the code to find the `OPERATOR_STS_PVC_TEMPLATE_NAME` variable in order to not delete our PVCs while upgrading.

It would help to make this variable "Official" in the docs, or add some other way to configure the name.

Thank you in advance 😃

vshulcz:
**Description**

Fixes #1365.

`OPERATOR_STS_PVC_TEMPLATE_NAME` is read in `statefulset.go` and `finalizer.go` but is not documented, so there is no way to find it other than reading the source.

The volume claim template defaults to the StatefulSet name, so PVCs come out as `redis-redis-0` for a standalone `Redis` named `redis`, and `redis-cluster-leader-redis-cluster-leader-0` for a `RedisCluster`. When PVCs were created under a different template name by an older version, the new StatefulSet does not adopt them and Redis comes up on an empty volume. This adds a short section to the operator upgrade page explaining that, showing how to set the variable on the operator deployment, and noting that it applies to every managed setup.

**Type of change**

* Documentation

### DIFF
--- a/docs/content/en/docs/Installation/Upgradation/_index.md
+++ b/docs/content/en/docs/Installation/Upgradation/_index.md
@@ -54,6 +54,24 @@ annotations:
   redisSentinel.opstreelabs.in/skip-reconcile: "true"
 ```
 
+## Preserving existing PVC names
+
+The name of the volume claim template in the generated StatefulSet defaults to the StatefulSet name, which gives PVCs such as `redis-redis-0` for a standalone `Redis` named `redis`, and `redis-cluster-leader-redis-cluster-leader-0` for a `RedisCluster`. If an older operator version created your PVCs under a different template name, a new StatefulSet will not adopt them and Redis starts on an empty volume.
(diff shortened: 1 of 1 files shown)


shubham-cmyk (approved):
LGTM!

PR #1869, opened 2026-08-20
fix(chart): allow configuring TLS secret volume mode and default to 0440

### LINKED ISSUE #1868: redis-cluster chart drops every TLS.secret field except secretName

redis-operator version: v0.26.0
redis-cluster chart version: 0.17.4

Yes, and on `main` too. `charts/redis-cluster/templates/redis-cluster.yaml` is unchanged there.

$ kubectl version --client
Client Version: v1.36.3
Kustomize Version: v5.8.1

# darwin/arm64, helm v4.2.4
</pre></details>

No cluster is needed to reproduce this, `helm template` is enough.

**What did you do?**

The `RedisCluster` CRD declares `spec.TLS.secret` as a full `corev1.SecretVolumeSource`, and the operator hands it to the StatefulSet as-is:

```go
// internal/k8sutils/statefulset.go
if containerParams.TLSConfig != nil {
    statefulset.Spec.Template.Spec.Volumes = append(statefulset.Spec.Template.Spec.Volumes,
        corev1.Volume{
            Name: "tls-certs",
            VolumeSource: corev1.VolumeSource{
                Secret: &containerParams.TLSConfig.Secret,
            },
        })
}
```

So `defaultMode`, `items` and `optional` all work if you apply the CR directly. The chart copies out only the name:

```yaml
# charts/redis-cluster/templates/redis-cluster.yaml
    secret:
      secretName: {{ .Values.TLS.secret.secretName | quote }}
```

Reproduce:

```shell
helm template t charts/redis-cluster --set TLS.secret.secretName=demo-tls --set TLS.secret.defaultMode=256 | grep -A 8 "TLS:"
```

```yaml
  TLS:
    ca: "ca.crt"
    cert: "tls.crt"
    key: "tls.key"
    secret:
      defaultMode: 256
      secretName: "demo-tls"
```

**What did you see instead?**

```yaml
  TLS:
    ca: "ca.crt"
    cert: "tls.crt"
    key: "tls.key"
    secret:
      secretName: "demo-tls"
```

`defaultMode` is dropped silently. The practical effect is that the Redis TLS key material is mounted with the Kubernetes default of 0644, so the private key is readable by any user on the node at `/var/lib/kubelet/pods/<uid>/volumes/kubernetes.io~secret/tls-certs/..data/tls.key`, and there is no way to tighten it from chart values.

On the mode itself: secret volumes are always mounted read-only, so kubelet ORs `0440` into whatever is requested (`pkg/volume/volume_linux.go`). That makes `0440` the tightest reachable mode, and it works with the chart's default `podSecurityContext` (`runAsUser: 1000`, `fsGroup: 1000`) because kubelet chowns the files to `root:fsGroup`.

mrcat71:
Closes #1868

### What

`charts/redis-cluster` copied a single field out of `TLS.secret` and discarded the rest:

```yaml
    secret:
      secretName: {{ .Values.TLS.secret.secretName | quote }}
```

The `RedisCluster` CRD declares `spec.TLS.secret` as a full `corev1.SecretVolumeSource`, and the operator forwards it to the StatefulSet untouched, so `defaultMode`, `items` and `optional` already work when the CR is applied directly. Only the chart was dropping them.

This passes the whole map through instead:

```yaml
    secret:
      {{- toYaml .Values.TLS.secret | nindent 6 }}
```

and sets `defaultMode: 0440` as the chart default in `values.yaml`.

### Why 0440

~~Kubernetes mounts secret volumes at 0644 unless told otherwise, which leaves the Redis private key readable by every user on the node at `/var/lib/kubelet/pods/<uid>/volumes/kubernetes.io~secret/tls-certs/..data/tls.key`.~~

Secret volumes are always mounted read-only, so kubelet ORs `0440` into whatever mode is requested (`pkg/volume/volume_linux.go`). `0440` is therefore the tightest mode actually reachable, and asking for anything stricter such as `0400` still yields `0440`.

It works with the chart's default `podSecurityContext` (`runAsUser: 1000`, `fsGroup: 1000`) because kubelet chowns the files to `root:fsGroup` and redis-server reads them through the group bit.

### Compatibility

The template change alone is backwards compatible. Values that only set `secretName` render exactly as before apart from quoting, which is YAML-equivalent.

The `defaultMode: 0440` default is the part with a behaviour change. It is safe for the chart's own defaults, but a user who overrides `podSecurityContext` to drop `fsGroup` while running as non-root would find redis-server unable to read its own key. Say the word and I will drop that hunk and keep only the passthrough, which still lets operators set the mode themselves.

### Verification

```
helm lint charts/redis-cluster            # 0 failed
```

TLS enabled renders the mode, TLS disabled is unchanged:

```
$ helm template t charts/redis-cluster --set TLS.secret.secretName=demo-tls | grep -A 6 "^  TLS:"
  TLS:
    ca: "ca.crt"
    cert: "tls.crt"
    key: "tls.key"
    secret:
      defaultMode: 288
      secretName: demo-tls
```

`288` is `0440` in decimal, which is what the int32 field expects.

Chart version and appVersion bumped to 0.17.5, and `README.md` regenerated with `helm-docs`.

One note for reviewers: `--set TLS.secret.defaultMode=0400` renders the string `"0400"` and would fail CRD validation. That is standard Helm `--set` behaviour for leading-zero numerics rather than anything specific to this chart. A values file, or `--set ...=256`, both produce a correct integer.

### DIFF
--- a/charts/redis-cluster/values.yaml
+++ b/charts/redis-cluster/values.yaml
@@ -257,12 +257,6 @@ TLS:
   # RedisCluster spec as-is.
   secret:
     secretName: ""
-    # Kubernetes mounts secret volumes with mode 0644 unless told otherwise,
-    # which leaves the Redis private key readable by every user on the node.
-    # Secret volumes are always mounted read-only, so kubelet ORs 0440 into
-    # whatever is set here, making 0440 the tightest reachable mode. It works
-    # with the podSecurityContext above because kubelet chowns the files to
-    # root:fsGroup.
     defaultMode: 0440
 
 acl:
--- a/charts/redis-cluster/values.yaml
+++ b/charts/redis-cluster/values.yaml
@@ -253,8 +253,17 @@ TLS:
   ca: ca.crt
   cert: tls.crt
   key: tls.key
+  # Any corev1.SecretVolumeSource field is accepted here and passed to the
+  # RedisCluster spec as-is.
   secret:
     secretName: ""
+    # Kubernetes mounts secret volumes with mode 0644 unless told otherwise,
+    # which leaves the Redis private key readable by every user on the node.
+    # Secret volumes are always mounted read-only, so kubelet ORs 0440 into
+    # whatever is set here, making 0440 the tightest reachable mode. It works
+    # with the podSecurityContext above because kubelet chowns the files to
+    # root:fsGroup.
+    defaultMode: 0440
 
 acl:
   # Configure ACL from Kubernetes Secret
--- a/charts/redis-cluster/Chart.yaml
+++ b/charts/redis-cluster/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v2
 name: redis-cluster
 description: Provides easy redis setup definitions for Kubernetes services, and deployment.
-version: 0.17.4
-appVersion: "0.17.4"
+version: 0.17.5
+appVersion: "0.17.5"
 home: https://github.com/ot-container-kit/redis-operator
 sources:
   - https://github.com/ot-container-kit/redis-operator
--- a/charts/redis-cluster/README.md
+++ b/charts/redis-cluster/README.md
@@ -49,6 +49,7 @@ helm delete <my-release> --namespace <namespace>
 | TLS.ca | string | `"ca.crt"` |  |
 | TLS.cert | string | `"tls.crt"` |  |
 | TLS.key | string | `"tls.key"` |  |
+| TLS.secret.defaultMode | int | `288` |  |
 | TLS.secret.secretName | string | `""` |  |
 | acl.secret.secretName | string | `""` |  |
 | annotations | object | `{}` |  |
--- a/charts/redis-cluster/templates/redis-cluster.yaml
+++ b/charts/redis-cluster/templates/redis-cluster.yaml
@@ -94,7 +94,7 @@ spec:
     cert: {{ .Values.TLS.cert | quote }}
     key: {{ .Values.TLS.key | quote }}
     secret:
-      secretName: {{ .Values.TLS.secret.secretName | quote }}
+      {{- toYaml .Values.TLS.secret | nindent 6 }}
   {{- end }}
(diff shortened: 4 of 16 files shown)


mrcat71:
Extended to the remaining charts, since redis, redis-replication and redis-sentinel had the identical secretName-only copy and the same podSecurityContext defaults. Leaving three known instances of the same bug behind did not seem useful.

mrcat71:
Looks like the error in https://github.com/OT-CONTAINER-KIT/redis-operator/actions/runs/32486648358/job/96787975845?pr=1869 isn't related with this PR changes

shubham-cmyk:
@mrcat71 let me check if the every use can read under the `/var/lib/kubelet`

shubham-cmyk:
@mrcat71 i have checked the auth it's like this 

```
/var/lib/kubelet # stat -c '%a %U:%G' /var/lib/kubelet
755 root:root
/var/lib/kubelet # 
/var/lib/kubelet # stat -c '%a %U:%G' /var/lib/kubelet/pods
750 root:root
/var/lib/kubelet # ls

```

it seems to be non privileged user can't access the pods in the kubelet, so secrets would be safe we can change the description of this PR about this, PR looks good we can give this option to configure anyways

charts/redis-cluster/values.yaml
```diff
@@ -253,8 +253,17 @@ TLS:
   ca: ca.crt
   cert: tls.crt
   key: tls.key
+  # Any corev1.SecretVolumeSource field is accepted here and passed to the
+  # RedisCluster spec as-is.
   secret:
     secretName: ""
+    # Kubernetes mounts secret volumes with mode 0644 unless told otherwise,
+    # which leaves the Redis private key readable by every user on the node.
```
shubham-cmyk: remove this comments since only root can access

mrcat71:
@shubham-cmyk agreed, comment removed

shubham-cmyk (approved):

Issue #1347, opened 2025-05-11
Service updated before Statefulset during Reconcilation


jpadmin:
**What version of redis operator are you using?**

```shell
kubectl logs <_redis-operator_pod_name> -n <namespace>
```

redis-operator version: 0.20.1

**Does this issue reproduce with the latest release?**

Yeah, this reported version is the latest as of now.

**What operating system and processor architecture are you using (`kubectl version`)?**

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version
Client Version: v1.29.3
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.31.4+k3s1
</pre></details>

**What did you do?**

Please follow the below steps to reproduce the problem.

1. Create a redis cluster object with minimal settings with atleast 1 leader and follower but without setting the annotation - `redis.opstreelabs.in/recreate-statefulset: "true"`. For eg use [this](https://github.com/OT-CONTAINER-KIT/redis-operator/blob/main/example/v1beta2/redis-cluster.yaml) manifest.
2. Wait for the cluster to become ready.
3. Update the redis cluster object and add a dummy label.

**What did you expect to see?**

Operator should stop reconcilation without making any changes to the underlying kubernetes manifests - especially stateful sets and service and flag the below error.
```
Forbidden: updates to statefulset spec for fields other than 'replicas', 'ordinals', 'template', 'updateStrategy', 'persistentVolumeClaimRetentionPolicy' and 'minReadySeconds' are forbidden
```
However, this is not propagated to follower services because the success of the leader statefulset patch update is actually acting as a deciding factor for the follower changes.

**What did you see instead?**
Operator update the labels for the leader service and then only stop reconcilation stating that the statefulset update is forbidden. This results in the leader service to become completely unavailable due to mismatch in the leader pod labels and leader service selector labels.

**What can be done as an immediate step to avoid the problem?**

Switching the order of leader statefulset and service updates during reconciliation should fix the problem(but not sure about any other test cases other than cluster creation and label update).

**What is a good approach to address these problems?**
- Implement a validation mechanism that decides what objects are updated based on the change-sets during the reconcile step.
- Try to pick a definitive set of labels for the service selectors instead of picking all the labels that are applied to the rediscluster object.

dxygit1:
The latest image I'm using, quay.io/opstree/redis-operator:v0.23.0, also has this problem. Is there an updated image for this?

[...]

PR #1875, opened 2026-08-21
fix(cluster): create NodePort services before StatefulSets

cuongquochoang:
## What this PR does

- Creates missing per-pod NodePort Services before rendering leader and follower StatefulSets, so Kubernetes-assigned client and bus ports are present in the Redis announce environment on the first revision.
- Leaves existing Services untouched during the preflight pass. Normal Service reconciliation still runs after the StatefulSet, preserving the StatefulSet-first update safety introduced for #1347.
- Covers initial creation, scale-up, non-NodePort clusters, and Service lookup failures with regression tests.

## Why

The Redis Cluster StatefulSet reads allocated NodePorts from the per-pod Services to populate `announce_port_*` and `announce_bus_port_*`. Previously the StatefulSet was created before those Services, so its first revision omitted the variables and Redis exited because `cluster-announce-port` had no value. A later reconciliation updated the StatefulSet template, but the CrashLooping pods were not automatically replaced.

The same ordering issue also occurred when scaling up because the StatefulSet existed while the new ordinal Service did not.

## Validation

- `KUBEBUILDER_ASSETS=<envtest-assets> go test ./... -coverprofile cover.out`
- `golangci-lint run` (0 issues)
- `make manager`
- Live Kubernetes test with a 3-leader/3-follower NodePort RedisCluster: both StatefulSets converged on revision 1, all six pods became Ready with 0 restarts, and Redis reported `cluster_state:ok`, 6 known nodes, and 3 shards.

### DIFF
--- a/internal/controller/rediscluster/rediscluster_controller.go
+++ b/internal/controller/rediscluster/rediscluster_controller.go
@@ -169,6 +169,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		}
 	}
 
+	err = k8sutils.EnsureRedisClusterNodePortServices(ctx, instance, "leader", r.K8sClient)
+	if err != nil {
+		return intctrlutil.RequeueE(ctx, err, "")
+	}
 	err = k8sutils.CreateRedisLeader(ctx, instance, r.K8sClient)
 	if err != nil {
 		return intctrlutil.RequeueE(ctx, err, "")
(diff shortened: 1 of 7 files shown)


quoccuongxXx:
Pushed 6c875f5, a follow-up that closes gaps I found while re-reviewing my own change. The preflight fixed the ordering, but it did not make the render itself safe.

**The render swallowed Service lookup errors.** `generateRedisClusterContainerParams` logged the error from `getService` and carried on, so a transient API error produced a pod template without the announce variables — the same broken StatefulSet revision the preflight was meant to prevent, just triggered by a failed read instead of a missing Service. It now returns the error and `CreateRedisClusterSetup` aborts before touching the StatefulSet.

**`Spec.Ports[0]` / `Spec.Ports[1]` could panic the operator.** A per-pod Service without the cluster-bus port — for example one created before #1406, or edited by hand — makes the index panic and takes down the manager. This mattered more after my change: the follower used to run `CreateRedisFollowerService` before the render, which repaired a drifted Service in the same loop, and the preflight deliberately skips Services that already exist. Ports are now matched by name, and an unallocated node port is rejected rather than announced as `0`.

**The env slice was aliased with `Spec.EnvVars`.** `util.Coalesce` returns the pointer unchanged, so appending wrote straight into the cluster spec, which the leader and the follower render share within one reconciliation. Whenever `spec.env` was set, the follower pod template inherited the leader's `announce_port_*` variables plus a duplicate `NODEPORT`, and follower pods rolled whenever a leader node port changed. The render now starts from a copy.

**The preflight can no longer update a live Service.** It went through `CreateOrUpdateService`, which does its own lookup and falls through to `patchService` if the Service exists by then — exactly the selector-ahead-of-StatefulSet regression from #1347/#1348 that the preflight is supposed to preserve. It now builds the definition and calls `createService` directly, treating `AlreadyExists` as success.

**Regression coverage.** The unit tests only covered the new helper against a fake clientset, which never allocates node ports, so nothing verified the actual fix. There is now an envtest spec (`testdata/nodeport.yaml`) asserting that the first StatefulSet revision of both roles carries announce variables equal to the node ports the API server allocated, that the generation stays at 1, and that the leader variables do not leak into the follower template.

I verified the new spec fails without the fixes:

- against `origin/main` the leader template has no announce variables at all, and the logs show them arriving in a second StatefulSet update
- against bc97fea the follower template carries 12 announce variables instead of 6, the leader's six having leaked through the shared spec slice

`go build ./...`, the full `go test ./...` with the 1.31 envtest assets, `golangci-lint run` (0 issues) and `git diff --check` are all clean.

chungtd203338:
LGTM

PR #1876, opened 2026-08-22
perf(test): cut e2e wall clock from ~44min

drivebyer:
## Why

The E2E job took ~44 minutes. On [run 32485942320](https://github.com/OT-CONTAINER-KIT/redis-operator/actions/runs/32485942320) (all green) it was 44m27s, of which `chainsaw test` was 40m16s.

Two numbers set the direction:

- The 23 cases summed to **4783s**. At `parallel: 2` the floor is 2391s and the run took 2416s — the suite was already packed to **99%** of what that parallelism allows, so raising `parallel` alone would have bought nothing.
- **`setup-redis-cluster` alone was 1815s**, 75% of the wall clock, and no amount of parallelism gets below the longest single case.

So the work had to go into what each case costs.

## Results

Measured on this branch:

| | before | after |
|---|---|---|
| `chainsaw test` wall clock | 40m16s | **19m24s (−52%)** |
| E2E job | 44m27s | **23m11s** |
| `setup-redis-cluster` | 1816s | 1005s (−45%) |
| `setup-redis-sentinel` | 333s | **33s (−90%)** |
| `acl-user-redis-cluster` | 164s | 93s (−43%) |
| `pvc-shrink-guard-redis-standalone` | 70s | 41s (−41%) |
| `redis-cluster-restart` | 479s | 315s (−34%) |
| `setup-ha` | 316s | 215s (−32%) |

On top of that the job now starts ~12 minutes earlier in the run, so PR wall clock improves by more than the suite numbers alone show.

### Why `parallel: 3` and not 4

Per-case cost and wall clock pull in opposite directions once cases run concurrently. Measured at both settings:

| | `parallel: 2` | `parallel: 3` |
|---|---|---|
| sum of all 23 cases | 2915s | 3289s (+13% contention tax) |
| `setup-redis-cluster` | 871s | 1005s |
| wall clock | 24m44s | **19m24s** |

3 is where it stops paying. The wall clock is already within 16% of the 1005s critical path, so `sum/4 = 822s` is below it — a fourth lane cannot shorten the run, it can only lengthen the longest case by adding contention. The model puts 4 at ~1206s, i.e. worse than the 1164s measured here.

Getting past this needs `setup-redis-cluster` split into independent cases, not more lanes.

## What changed

**Gate `e2e_test` on `lint` alone.** It hung off `validate_yaml` and `helm_chart_test`, which pull in the whole `gotest` chain. The job did not start until 12.5 minutes into the run, and neither dependency tests anything the suite relies on.

**Drop `cleanup.skipDelete`.** The old log shows 24 × `CLEANUP | SKIP` — every namespace and its pods stayed on the cluster for the rest of the run. By the time `setup-redis-cluster` reached its scale-out there were 75+ redis pods on a 4-vCPU runner, which is why the same operations cost 127s early in a run and ten minutes late in it. It was added alongside `parallel: 10` in #1114 to dodge cleanup timeouts; the real fix for those is `timeouts.cleanup` (below), not skipping the delete.

**Take kind down to 1 control-plane + 3 workers.** kind keeps the `NoSchedule` taint on control-planes whenever a cluster has workers, so the two extra control-planes added a second and third etcd/apiserver/controller-manager plus an haproxy front-end and contributed *zero* schedulable capacity. They came from #1237 for node-shutdown simulation and pod anti-affinity; the suite has neither today. Cluster creation went 74s → ~50s.

**Preload test images and switch the manifests to `IfNotPresent`.** 41 `imagePullPolicy: Always` against one `IfNotPresent`, and kind nodes do not share an image store, so the same image was pulled from quay.io once per pod per node. This also pins a run to a single digest of the moving `:latest` tags. The preload is best-effort — a failed pull or side-load warns and lets the kubelet pull on demand.

**Drop `limits.cpu` from the test CRs**, and from the operator for the test run only (`config/manager` keeps its shipped default). `redis-cli --cluster reshard` is exec'd *inside* the redis pod (`executeCommand1` in `internal/k8sutils/redis.go`), so it shared the container's quota with redis-server — and that quota was `101m`, a tenth of a core. That is what the 672s and 481s scale-out/scale-in steps were. Memory limits are untouched: `maxMemoryPercentOfLimit` derives `maxmemory` from them and `max-memory-env` asserts the resulting `83886080`.

**Replace seven fixed `sleep` steps (690s) with polls.** `setup/redis-sentinel` was the worst — `sleep 5m` inside a 333s case, and it was the only case that did not move at all when the CPU limits came off. The polls wait on the thing each step actually wants: the pod label after a cluster failover, `status.masterNode` after a replication failover, the services answering, the `storageCapacity` annotation after an expand.

Two sleeps stay, and now say why in a comment: `pvc-shrink-guard` and `pod-management-policy` each prove a *non-event* (the shrink is ignored, the immutable field is not applied), and there is nothing to poll for.

**Drop `delayBeforeCleanup: 10s`.** Free while `skipDelete` was set; with cleanup back on it is 10s per case — +100% on an 11s case, 230s of the total.

**Set `timeouts.cleanup: 2m`.** `sentinel-pod-placement` failed namespace teardown with `context deadline exceeded` 30s after the delete: the CRD finalizers need longer than chainsaw's 30s default. Worth knowing — chainsaw records that as a failed test in its go-test output but **not** in its summary or exit code, so that run reported green with `--- FAIL: chainsaw/sentinel-pod-placement` sitting in the log.

**Make `cluster-scale-out.yaml` differ from `cluster.yaml` only in `clusterSize`.** It omitted `TLS`, `redisConfig` and the service block, so it read as though scaling out tore those down. It did not — chainsaw applies with a merge patch, so the omitted fields survived. No behaviour change; the file now says what it does rather than relying on merge-patch semantics.

## Two failures this surfaced, both pre-existing

**data-assert OOM.** `setup-redis-cluster` died on `exit code 137` while `go run` was downloading modules: two data-assert pods were compiling Go at the same moment and one ran past its 512Mi limit. Given headroom (1Gi) and `GOFLAGS=-p=2` so peak RSS does not scale with the node's core count.

**A racy replication assertion.** `acl-pvc-redis-replication` wrote to the master and read from the replica 30s later. `ready-replication.yaml` only asserts `status.masterNode` — that the operator picked a master, not that the replica is attached and synced. The fixed sleep was covering that gap by accident, and with images preloaded and the CPU limits gone the pods reach Ready sooner. It now polls; in practice the replica needed **more than the 30s** the test used to allow, so the old version was genuinely racy.

## Verification

- `yamllint --strict ./tests/` and `actionlint` on the workflow: no new findings.
- Brought the new kind config up for real: 4 nodes in 21s locally, control-plane still carrying `node-role.kubernetes.io/control-plane:NoSchedule`.
- Server-side dry-run applied all 27 modified CR manifests against a kind cluster with the CRDs installed. (The two `pvc-shrink-guard` patch fragments fail standalone validation both before and after; they are partial documents meant to be merged onto an existing CR.)
- All 8 new polling scripts pass `sh -n` — chainsaw runs `script` under `sh -c`, so they have to be POSIX — and were exercised against a stubbed `kubectl` for both the success and the timeout path.
- Checked that no assert file pins a container CPU value, so removing the limits cannot silently change what the suite verifies. 39 CPU limits removed, 39 memory limits still present.

## Not in this PR

- `data-assert` still compiles Go inside the cluster on every run, ~57s per first use across four cases. A prebuilt binary in a small image would remove that and the OOM class with it.
- `setup-redis-cluster` at 1005s is the critical path and caps any further parallelism; splitting it is the next structural step.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -212,11 +212,12 @@ jobs:
     uses: ./.github/workflows/call-test-charts.yaml
 
   e2e_test:
-    # DEBUG
-    needs: [validate_yaml, helm_chart_test]
+    # Only gate on lint. The suite used to hang off validate_yaml and
+    # helm_chart_test, which pulled in the whole gotest chain and delayed the
+    # job by ~12 minutes without testing anything the suite depends on.
+    needs: [lint]
     name: E2E Test
     runs-on: ubuntu-latest
-    if: always() && (needs.validate_yaml.result == 'success') && (needs.helm_chart_test.result == 'success' || needs.helm_chart_test.result == 'skipped')
     steps:
       - name: Checkout code
         uses: actions/checkout@v6
@@ -257,10 +258,49 @@ jobs:
           kubectl cluster-info --context kind-kind
           kind load docker-image redis-operator:e2e --name kind
 
+      # Kind nodes do not share an image store, so an imagePullPolicy of Always
+      # made every pod re-pull from quay.io on every node. Pull each image once
+      # here and side-load it into the nodes instead; the test manifests use
+      # IfNotPresent so nothing goes back over the network. This also pins the
+      # whole run to a single digest of the moving :latest tags.
+      - name: Preload test images into Kind
+        run: |
+          set -euo pipefail
+          IMAGES=(
+            quay.io/opstree/redis:latest
+            quay.io/opstree/redis:v7.0.15
+            quay.io/opstree/redis-sentinel:latest
+            quay.io/opstree/redis-exporter:v1.44.0
+            redis:alpine
+            busybox:latest
+            busybox:1.36
+          )
+          # Docker Hub rate-limits anonymous pulls from shared runner IPs, so a
+          # pull can fail for reasons that have nothing to do with this change.
+          printf '%s\n' "${IMAGES[@]}" \
+            | xargs -P 4 -n 1 -I{} sh -c 'docker pull -q "{}" || echo "::warning::pull failed for {}"'
+          # Side-loading is an optimisation, not a correctness requirement: the
+          # manifests ask for IfNotPresent, so anything that fails here is just
+          # pulled by the kubelet on first use. Warn rather than fail the job --
+          # `kind load` cannot import a multi-arch image saved from Docker's
+          # containerd image store, and that is not worth breaking E2E over.
+          for img in "${IMAGES[@]}"; do
+            kind load docker-image "$img" --name kind \
+              || echo "::warning::could not preload ${img}; kubelet will pull it on demand"
+          done
+
       - name: Install Redis Operator
         run: |
           make deploy IMG=redis-operator:e2e
           kubectl set env deployment/redis-operator-redis-operator -n redis-operator-system INIT_CONTAINER_IMAGE=redis-operator:e2e FEATURE_GATES=GenerateConfigInInitContainer=true MAX_CONCURRENT_RECONCILES=10
+          # config/manager ships a 100m CPU cap. With MAX_CONCURRENT_RECONCILES=10
+          # and the operator streaming every `redis-cli --cluster` exec itself,
+          # that cap throttles the suite rather than protecting anything. Lift it
+          # for the test run only -- the shipped default is left untouched.
+          # (`set resources --limits=cpu=0` would write cpu: "0", which then fails
+          # the requests<=limits check; a strategic merge with null drops the key.)
+          kubectl -n redis-operator-system patch deployment redis-operator-redis-operator \
+            -p '{"spec":{"template":{"spec":{"containers":[{"name":"manager","resources":{"limits":{"cpu":null}}}]}}}}'
 
       - name: Wait for Redis Operator to be ready
         run: |
--- a/tests/_config/chainsaw-configuration.yaml
+++ b/tests/_config/chainsaw-configuration.yaml
@@ -7,7 +7,12 @@ metadata:
 spec:
   execution:
     failFast: true
-    parallel: 2
+    # 3, not 4. With the suite at 2915s across 23 cases, sum/3 = 972s still sits
+    # above the 871s critical path (setup-redis-cluster), so there is slack for
+    # contention to eat. sum/4 = 729s puts the wall clock exactly on the
+    # critical path, where any slowdown in that one case shows up one-for-one.
+    # 4 is the ceiling worth trying -- 5 buys nothing.
+    parallel: 3
     forceTerminationGracePeriod: 5s
   # No cleanup block: both settings it used to carry are back at their defaults.
   #
@@ -21,6 +26,11 @@ spec:
   # 3763s. Two full runs finished with no cleanup timeouts, so the settling
   # time it bought is not needed.
   timeouts:
+    # Namespace teardown has to wait out the CRD finalizers, and chainsaw's
+    # 30s default is not enough: sentinel-pod-placement hit "context deadline
+    # exceeded" on it. Chainsaw counts that as a failed test in its go-test
+    # output but not in its summary or exit code, so CI went green over it.
+    cleanup: 2m
     apply: 5m
     delete: 5m
     assert: 20m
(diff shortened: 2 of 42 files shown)


shubham-cmyk:
@drivebyer don't merge this immediately i would also look in the above changes.

drivebyer:
@shubham-cmyk  cc

.github/workflows/ci.yaml
```diff
@@ -212,11 +212,12 @@ jobs:
     uses: ./.github/workflows/call-test-charts.yaml
 
   e2e_test:
-    # DEBUG
-    needs: [validate_yaml, helm_chart_test]
+    # Only gate on lint. The suite used to hang off validate_yaml and
```
shubham-cmyk, line 215: Why are we removing the `validate_yaml` before the e2e run

.github/workflows/ci.yaml
```diff
@@ -257,10 +258,49 @@ jobs:
...
+      # made every pod re-pull from quay.io on every node. Pull each image once
+      # here and side-load it into the nodes instead; the test manifests use
+      # IfNotPresent so nothing goes back over the network. This also pins the
+      # whole run to a single digest of the moving :latest tags.
+      - name: Preload test images into Kind
+        run: |
+          set -euo pipefail
+          IMAGES=(
+            quay.io/opstree/redis:latest
+            quay.io/opstree/redis:v7.0.15
```
shubham-cmyk, line 271: don't load this image i don't think it is getting used plus we probably don't want to touch CI if we release any new tag

.github/workflows/ci.yaml
```diff
@@ -257,10 +258,49 @@ jobs:
...
+            quay.io/opstree/redis-sentinel:latest
+            quay.io/opstree/redis-exporter:v1.44.0
```
shubham-cmyk, line 273: same goes with this

.github/workflows/ci.yaml
```diff
@@ -257,10 +258,49 @@ jobs:
           kubectl cluster-info --context kind-kind
           kind load docker-image redis-operator:e2e --name kind
 
+      # Kind nodes do not share an image store, so an imagePullPolicy of Always
+      # made every pod re-pull from quay.io on every node. Pull each image once
+      # here and side-load it into the nodes instead; the test manifests use
+      # IfNotPresent so nothing goes back over the network. This also pins the
+      # whole run to a single digest of the moving :latest tags.
+      - name: Preload test images into Kind
+        run: |
```
shubham-cmyk, line 267: can you keep this script of loading the image in the hack/ dir easier to maintain and read later

tests/_config/chainsaw-configuration.yaml
```diff
@@ -7,12 +7,32 @@ metadata:
...
+    # 3, not 4. With the suite at 2915s across 23 cases, sum/3 = 972s still sits
+    # above the 871s critical path (setup-redis-cluster), so there is slack for
+    # contention to eat. sum/4 = 729s puts the wall clock exactly on the
+    # critical path, where any slowdown in that one case shows up one-for-one.
+    # 4 is the ceiling worth trying -- 5 buys nothing.
+    parallel: 3
     forceTerminationGracePeriod: 5s
-  cleanup:
-    skipDelete: true
-    delayBeforeCleanup: 10s
```
shubham-cmyk, line 14: don't remove this, in case we want operator to update internal state of cluster etc before cleanup

tests/e2e-chainsaw/v1beta2/pvc-shrink-guard/redis-standalone/chainsaw-test.yaml
```diff
@@ -61,23 +62,24 @@ spec:
...
-    - name: Wait for operator to reconcile annotation
-      try:
-        - sleep:
-            duration: 30s
-
     - name: Verify annotation updated after successful expand
       try:
         - script:
-            timeout: 30s
+            timeout: 120s
```
shubham-cmyk, line 68:
```
    - name: Verify annotation updated after successful expand
      try:
        - assert:
            timeout: 120s
            resource:
              apiVersion: apps/v1
              kind: StatefulSet
              metadata:
                name: redis-standalone-shrink-test
                annotations:
                  # 2Gi = 2147483648 bytes
                  storageCapacity: "2147483648"
```


the assert already has the polling condition so we check like this instead of this script, replace this every where

Issue #1722, opened 2026-03-20
Redis cluster fails to recover after full restart – nodes.conf repair proposal


vincentmrg:
Hello,

First of all, thanks a lot for maintaining this operator, we are successfully using it in production and are very satisfied with it.

## **Is your feature request related to a problem? Please describe.**

When running Redis Cluster with the operator, we observed that the cluster does not recover automatically after a simultaneous restart of all nodes (e.g. full node pool or Availability Zone failure).

In this scenario:
- all Redis pods are restarted
- pods receive new IP addresses
- each node loads its persisted nodes.conf
- nodes attempt to reconnect using stale IPs
- several nodes are marked fail / fail?
- the cluster remains in a degraded state and does not self-heal

## **Describe the solution you'd like**

We would like to explore adding (or supporting) a mechanism to allow automatic recovery after full cluster restart, for example:
👉 A nodes.conf repair mechanism at startup, which would:
- resolve peer nodes via Kubernetes DNS (headless service)
- detect IP changes
- update nodes.conf accordingly before Redis starts

Tests have been conducted, detailed in the [Additional context](#additional_context) section, and the [Custom implementation: nodes.conf repair mechanism with Opstree Redis Operator](#custom_implem) solution appears to be functional.

Based on our experiments, this approach appears to work well, but we’d like your perspective on it.
- Does this solution seem sound from a Redis cluster perspective?
- Do you have any better ideas for solving the problem?
- Would this kind of mechanism make sense as an addition to the operator?

## **What version of redis-operator are you using?**

redis-operator version: [v0.24.0](https://github.com/OT-CONTAINER-KIT/redis-operator/releases/tag/v0.24.0)
redis-version: `quay.io/opstree/redis:v8.4.2`

## <a name="additional_context"></a>**Additional context**

We conducted several experiments to evaluate different approaches for handling full cluster restarts (e.g. AZ failure). Below is a summary of the approaches tested and their outcomes.

### Baseline behavior (Opstree Redis Operator)
We deployed a standard Redis Cluster (3 masters / 3 replicas) using the Opstree Redis Operator, with persistent nodes.conf.
We simulated a full infrastructure failure by terminating all nodes hosting Redis pods simultaneously.
Conclusion:
❌ The cluster does not recover automatically.
- Nodes restart with new IPs
- `nodes.conf` still contains old IPs
- Nodes cannot reach each other
- Cluster remains in `fail` / `fail?` state
[...]

PR #1877, opened 2026-08-22
fix(cluster): resolve cluster nodes that announce no hostname

cuongquochoang:
## What is broken

`getHostFromClusterNode` reads the address field of a `CLUSTER NODES` entry, splits it on `,` and requires exactly two elements:

```go
addressAndHost := node[1]
s := strings.Split(addressAndHost, ",")
if len(s) != 2 {
    return "", fmt.Errorf("failed to extract host from host and address string, unexpected number of elements: %d", len(s))
}
```

Redis formats that field as `<ip:port@cport[,hostname[,auxiliary-field=value]*]>`. The `,hostname` part is present only when the node runs with `cluster-announce-hostname` set.

A NodePort cluster does not set it. Every entry is a bare `<ip:port@cport>`, so `len(s) == 1` and the parse fails for every node. That has two consequences:

- `RepairStaleReplication` resolves no follower, repairs nothing, and logs `failed to repair stale replication links` on every reconciliation.
- `RepairDisconnectedNodes` fails the same way, silently, because it only runs once a node is already failed or disconnected. Self-healing is unavailable on exactly the clusters that need it.

The strict length check is also wrong in the other direction: Redis 7.2 and later may append auxiliary fields after the hostname, giving three or more elements, which is rejected just the same.

### Observed on a live cluster

A three-shard NodePort cluster on Kubernetes 1.35, Redis 8.4.2:

```text
$ redis-cli config get 'cluster-announce*'
cluster-announce-ip         192.168.10.16
cluster-announce-port       30445
cluster-announce-bus-port   30199
cluster-announce-hostname   <empty>

$ redis-cli cluster nodes
745ef44f91c7... 192.168.10.16:32253@32232 slave 9477a6bb... 0 ... connected
                └────────── node[1], no comma ──────────┘
```

The operator logged the error 81 times in 15 minutes on an otherwise healthy cluster.

## History

The logic arrived in de6b066 (#1105) as `getMasterHostFromClusterNode`, byte-identical. Its only caller was gated on masters that were already failed, so on a healthy cluster it never ran and the bug stayed latent.

1e7e76e (#1705) renamed it and added `repairStaleReplication`, which calls it for **healthy slaves**. That is what made it fire continuously. #1705 did not introduce the bug, it exposed it.

## The change

Parse the address field into its parts, then resolve the pod:

1. Through the announced hostname when Redis reports one. Unchanged behaviour for clusters that announce hostnames, and no additional API calls — the fallback index is built lazily.
2. Otherwise by matching the announced address against the addresses the operator expects each pod to announce: the node port of the pod's own Service for NodePort clusters, the pod IP for every other service type.

Matching a NodePort cluster on the **node port** rather than the node IP is deliberate. The node port identifies the pod on its own, and unlike the node IP it survives the pod being rescheduled onto another node, so a node recorded against a stale address still resolves. This mirrors `getEndpoint`, which already builds the same address in the forward direction.

A node the cluster has lost the address of is reported as `:0@0` with the `noaddr` flag. It cannot be matched by address at all, so it now says that plainly rather than reporting a failed lookup. Resolving those by node identity is a separate gap and is not addressed here.

## Tests

`TestRepairStaleReplication_nodePortClusterWithoutHostname` reproduces the reported failure end to end. Against `main` it fails with the production error:

```text
=== RUN   TestRepairStaleReplication_nodePortClusterWithoutHostname
    Error:  Received unexpected error:
            failed to extract host from host and address string, unexpected number of elements: 1
--- FAIL: TestRepairStaleReplication_nodePortClusterWithoutHostname (0.00s)
```

Also added: `TestParseClusterNodeAddress` covering an announced hostname, a bare address, auxiliary fields after the hostname, IPv6, a missing port and a truncated entry; and `TestClusterNodePodNamer` covering NodePort resolution, resolution of a node recorded against a stale node IP, pod-IP resolution for non-NodePort clusters, the hostname shortcut making no API call, the `noaddr` node, and the index being built once.

The existing `TestRepairDisconnectedNodes*` and `TestRepairStaleReplication_*` tests are unchanged in behaviour and still pass.

## Verification

`go build ./...`, `go test ./...` with the 1.31 envtest assets, `golangci-lint run` (0 issues) and `git diff --check` are all clean.

Deployed to the live cluster above. The error count went from 81 in 15 minutes to 0. The only failures that remained were for a genuinely address-less `:0@0` node, produced deliberately by destroying a pod's node identity.

### DIFF
--- a/internal/k8sutils/redis.go
+++ b/internal/k8sutils/redis.go
@@ -248,18 +248,18 @@ func repairDisconnectedNodes(ctx context.Context, client kubernetes.Interface, c
 	if err != nil {
 		return err
 	}
+	namer := newClusterNodePodNamer(client, cr)
 	var lastError error
 	for _, node := range nodes {
 		if !nodeFailedOrDisconnected(node) {
 			continue
 		}
-		host, err := getHostFromClusterNode(node)
+		podName, err := namer.podName(ctx, node)
 		if err != nil {
 			lastError = err
 			log.FromContext(ctx).V(1).Error(err, "Failed to get pod name from cluster node. Continuing with other nodes.", "Node", node)
 			continue
 		}
-		podName := strings.Split(host, ".")[0]
 		ip := getRedisServerIP(ctx, client, RedisDetails{
 			PodName:   podName,
 			Namespace: cr.Namespace,
@@ -303,12 +303,13 @@ func repairDisconnectedNodes(ctx context.Context, client kubernetes.Interface, c
 func RepairStaleReplication(ctx context.Context, client kubernetes.Interface, cr *rcvb2.RedisCluster) (int, error) {
 	redisClient := configureRedisClient(ctx, client, cr, cr.Name+"-leader-0")
 	defer redisClient.Close()
+	namer := newClusterNodePodNamer(client, cr)
 	return repairStaleReplication(ctx, redisClient, func(podName string) *redis.Client {
 		return configureRedisClient(ctx, client, cr, podName)
-	})
+	}, namer.podName)
 }
 
-func repairStaleReplication(ctx context.Context, redisClient *redis.Client, makeClient func(podName string) *redis.Client) (int, error) {
+func repairStaleReplication(ctx context.Context, redisClient *redis.Client, makeClient func(podName string) *redis.Client, podNameOf func(context.Context, clusterNodesResponse) (string, error)) (int, error) {
 	logger := log.FromContext(ctx)
 
 	nodes, err := clusterNodes(ctx, redisClient)
@@ -325,12 +326,11 @@ func repairStaleReplication(ctx context.Context, redisClient *redis.Client, make
 		if nodeFailedOrDisconnected(node) {
 			continue
 		}
-		host, err := getHostFromClusterNode(node)
+		podName, err := podNameOf(ctx, node)
 		if err != nil {
 			lastError = err
 			continue
 		}
-		podName := strings.Split(host, ".")[0]
 		masterNodeID := node[3]
 
 		followerClient := makeClient(podName)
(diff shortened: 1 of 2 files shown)


trynocoding:
Independent confirmation, plus two things this PR's description doesn't yet cover: the same parse failure on a **non-NodePort cluster that announces plain pod IPs**, and an end-to-end verification that the operations this repair path issues are exactly what converges such a cluster after a full restart. With the parse failure in place, the cluster stayed in `fail` for ~2.5 hours, because `RepairDisconnectedNodes` is the only actor that can break the gossip deadlock and it never got past the address parse.

### Environment

- redis-operator **v0.26.0** (`quay.io/opstree/redis-operator:v0.26.0`), Redis **v8.0.6** (`quay.io/opstree/redis:v8.0.6`)
- `RedisCluster`, 3 leaders / 3 followers, default (headless ClusterIP) services, **no `cluster-announce-hostname`**, `cluster-preferred-endpoint-type ip`
- `cluster-announce-ip` resolves to each pod's current IP at runtime (set by the image entrypoint, not pinned in redis.conf)
- single Kubernetes node; `/node-conf/nodes.conf` persisted on a PVC

### Trigger

Host machine reboot at `2026-09-09T01:28:21Z`. All six containers were killed in the same second (`lastState.terminated: reason=Unknown, exitCode=255`) and came back with an entirely new set of pod IPs — CNI re-assigns addresses after the node returns.

### Observed cluster state

Gossip is fully deadlocked: each node reconnects to the addresses recorded in its persisted `nodes.conf` (all stale), so no bus connection ever comes up, so nobody can gossip the announce-ip that *would* carry the new addresses. Every node has every peer in `fail?`; the seed leader's view:

```text
522c86aa393a... 100.123.94.69:6379@16379 master,fail? - ... connected 5461-10922
9ce106121ce5... 100.123.94.68:6379@16379 slave,fail? 522c86aa393a... ...
6c24442db9c7... 100.123.94.83:6379@16379 slave,fail? 18859f5ae884... ...
76729ebdd7eb... 100.123.94.80:6379@16379 myself,master - ... connected 0-5460
18859f5ae884... 100.123.94.118:6379@16379 master,fail? - ... connected 10923-16383
addbb5b167d2... 100.123.94.109:6379@16379 slave,fail? 76729ebdd7eb... ...
```

(`100.123.94.69/.68/.83/.118/.109` are the pre-reboot addresses of the other five pods; none of them exist any more.)

- `cluster_state:fail`, `cluster_slots_pfail:10923` (the two failed masters' ranges), `cluster_known_nodes:6`, zero gossip throughput
- verified directly: every stale address times out, every current pod IP answers `PONG` — the cluster is one round of `CLUSTER MEET <current-ip>` away from converging on its own
- `kubectl get rediscluster` meanwhile reports Ready 3/3: CR status only tracks pod readiness, so nothing surfaces the outage

### Operator side

First error at `01:29:00Z`, 39 seconds after the containers died; it has re-looped on requeue backoff ever since — 24 occurrences of the parse error in the last hour, one per failed node per reconcile:

```json
{
 "ts": "2026-09-09T01:29:00Z",
 "msg": "Failed to get pod name from cluster node. Continuing with other nodes.",
 "error": "failed to extract host from host and address string, unexpected number of elements: 1",
 "Node": [
  "522c86aa393a15aed269a518f320e1c3970d9257",
  "100.123.94.69:6379@16379",
  "master,fail?",
  "-",
  "1788917324402",
  "1788917323479",
  "4",
  "connected",
  "5461-10922"
 ]
}
```

Matches the PR's analysis exactly: bare `<ip:port@cport>` (no announced hostname, no auxiliary fields on 8.0.6), `len(s) == 1`, both `RepairDisconnectedNodes` and `RepairStaleReplication` resolve nothing.

### Recovery verification (manual repair)

We kept the cluster in the failed state on purpose, then ran the operations this repair path exists to issue — by hand, from the seed leader (all idempotent):

**1. `CLUSTER MEET <current-ip> 6379` for each of the five `fail?` nodes.**

The first `cluster info` poll after the meets already showed `cluster_state:ok` (slots still converging), and all 16384 slots were `ok` a few seconds later. Every `fail?` flag cleared — gossip re-established itself across the whole table from a single reachable member.

**2. Gossip alone did not restore replication for two of the three followers.**

With `CLUSTER NODES` fully clean, `INFO replication` on the three replica-role pods still showed two links pinned to their masters' **pre-reboot** addresses:

```text
follower-0: master_host:100.123.94.91 master_link_status:down   # .91 = leader-0 before reboot
follower-2: master_host:100.123.94.85 master_link_status:up     # picked up new IP via gossip
leader-1  : master_host:100.123.94.69 master_link_status:down   # .69 = follower-1 before reboot
```

**3. `CLUSTER REPLICATE <master-node-id>` on the two stale links** — the other half of `RepairDisconnectedNodes` for slave nodes, and the job of `RepairStaleReplication` — re-resolved the master's current IP from the now-healthy node table:

```text
follower-0: master_host:100.123.94.80 master_link_status:up
follower-2: master_host:100.123.94.85 master_link_status:up
leader-1  : master_host:100.123.94.127 master_link_status:up
```

Final state: `cluster_state:ok`, 16384/16384 slots ok, zero fail flags, 3/3 replication links up.

### Why this datapoint matters

The PR's live-cluster evidence is a NodePort cluster and an *otherwise healthy* one, where the failure shows up as log noise. This reproduction is the worst case of the sentence "self-healing is unavailable on exactly the clusters that need it": after a simultaneous restart of all nodes (issue #1722's scenario) there is no surviving bus connection to carry the new announce-ip around, so an operator-issued `CLUSTER MEET` is the **only** recovery path, and the parse failure removes it entirely.

The manual repair above adds two concrete observations:

- The parse failure is the **only** thing standing between this deployment shape and a fully automatic recovery — the operations the repair path issues are precisely what converged the cluster by hand.
- `CLUSTER MEET` alone is not sufficient: two replication links stayed pinned to stale addresses after gossip had fully converged, and only recovered on an explicit `CLUSTER REPLICATE`. That matches the two-stage design here (`RepairDisconnectedNodes` for failed nodes, `RepairStaleReplication` for connected-but-stale followers) — both stages are needed for the full-restart case, once address resolution works.

The hostname-then-address-matching resolution proposed in this PR covers this deployment shape as well (pods announce their own IPs, ClusterIP/headless services), so from our side it fixes the full-restart case too.

PR #1886, opened 2026-09-03
fix(replication): restart replica pods when the password secret changes

### LINKED ISSUE #1717: Updated Secret is not Synced with Replication PODs
What version of redis operator are you using?

redis-operator version:
v0.21.2

What did you do?

Setup:

RedisReplication with clusterSize 3
RedisSentinel with clusterSize 3

Steps to reproduce:

Update redis secret but redis.conf still points to old secret value.

What did you expect to see?

Redis.conf with updated secret so sentinel.conf also gets updated and use new secret value to communicate with replication PODs.

What did you see instead?

K8s redis secret is updated with newer value but operator doesn't update redis.conf to refer to new secret value.

What seems to be the problem:

Secret is not synced with replication PODs.

Some logs:

redis.conf points to older password while k8 secret has newer password updated.

MacBook-Pro ~ % kubectl exec -it fboss-ccache-redis-replication-0 -n ot-redis --context infra -- sh
Defaulted container "fboss-ccache-redis-replication" out of: fboss-ccache-redis-replication, redis-exporter
/data $ cat /etc/redis/redis.conf 
bind 0.0.0.0 ::
tcp-backlog 511
timeout 0
tcp-keepalive 300
daemonize no
supervised no
pidfile /var/run/redis.pid
masterauth 3e01fc307823023fbb6b4d4b5c8cca58b5dd8827
requirepass 3e01fc307823023fbb6b4d4b5c8cca58b5dd8827
protected-mode yes
save 900 1
save 300 10
save 60 10000
appendonly yes
appendfilename "appendonly.aof"
dir /data
port 6379
include /etc/redis/external.conf.d/redis-additional.conf
Even sentinel has a reference of older password.

After I trigger rolling reboot for replication PODs, the new password is synced and redis.conf is updated on each replication POD and this also makes sentinel.conf refer to new password to communicate with replication PODs without restarting sentinel PODs.

Is this normal behavior? If yes, shouldn't secret be automatically synched with replication PODs just like it is synched with sentinel PODs after replication PODs updates the secret?

pujitha24:
**Description**

When the password Secret referenced by a `RedisReplication` CR is rotated, the replica pods keep authenticating with the old password until something manually restarts them. The password is wired into the pod as an environment variable via `SecretKeyRef`, and Kubernetes does not refresh an already-running container's env var when the referenced Secret's content changes, so `redis.conf`'s `masterauth`/`requirepass` stays stale indefinitely.

This adds the standard Kubernetes idiom for this problem: a sha256 checksum of the password Secret's data is computed on each reconcile and stamped onto the StatefulSet's pod template as an annotation (`redis.opstreelabs.in/password-secret-checksum`). When the Secret's content changes, the checksum changes, the pod template annotation changes, and Kubernetes performs a normal StatefulSet rolling restart to pick up the new password — no extra RBAC or Secret watch required, since the reconcile loop already reads the CR's namespace.

The fix is scoped to `RedisReplication` only, matching the issue. `RedisSentinel`/`RedisCluster`/standalone `Redis` are not touched.

Fixes #ISSUE

**Type of change**

**Checklist**

**Additional Context**

Validation performed:
- `go build ./...` passes.
- `go test ./internal/k8sutils/...` passes, including three new/changed tests: `Test_passwordSecretChecksum` (proves the checksum changes when the Secret's password rotates, and is empty for a missing secret or a nil client), `Test_generateRedisReplicationContainerParams_SecretChecksum` (proves the checksum is wired into the container params for `RedisReplication`), and `TestGenerateStatefulSetsDefSecretChecksumAnnotation` (proves the checksum lands on `Spec.Template.ObjectMeta.Annotations`, which is what actually triggers the StatefulSet rolling restart).
- `go test ./internal/controller/redisreplication/... -run 'TestReconcile|TestSentinel|TestBuild|TestConfigure'` (existing unit tests) passes unchanged.
- `golangci-lint run` (v2.4.0, pinned in the Makefile) reports 0 issues on the changed packages.
- Not run here: the repo's envtest-based `TestAPIs` suite and the chainsaw e2e suite, both of which need a local etcd/Kubernetes control plane not available in this sandbox. I confirmed via `git stash` that `TestAPIs` fails identically on unmodified `main` in this sandbox (missing `/usr/local/kubebuilder/bin/etcd`), so that failure is a pre-existing environment limitation, not something introduced by this change.
- No API/CRD types were changed, so no `make manifests` regeneration was needed.

Fixes #1717

### DIFF
--- a/internal/controller/common/constants.go
+++ b/internal/controller/common/constants.go
@@ -3,6 +3,7 @@ package common
 const (
 	AnnotationKeyRecreateStatefulset         = "redis.opstreelabs.in/recreate-statefulset"
 	AnnotationKeyRecreateStatefulsetStrategy = "redis.opstreelabs.in/recreate-statefulset-strategy"
+	AnnotationKeyPasswordSecretChecksum      = "redis.opstreelabs.in/password-secret-checksum"
 )
 
 const (
--- a/internal/k8sutils/redis-replication.go
+++ b/internal/k8sutils/redis-replication.go
@@ -2,6 +2,9 @@ package k8sutils
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"sort"
 
 	rrvb2 "github.com/OT-CONTAINER-KIT/redis-operator/api/redisreplication/v1beta2"
 	rsvb2 "github.com/OT-CONTAINER-KIT/redis-operator/api/redissentinel/v1beta2"
@@ -9,6 +12,7 @@ import (
 	"github.com/OT-CONTAINER-KIT/redis-operator/internal/util"
 	"github.com/OT-CONTAINER-KIT/redis-operator/internal/util/maps"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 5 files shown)


shubham-cmyk:
@pujitha24 Please add the e2e chainsaw test case to validate this change

pujitha24:
@shubham-cmyk added a chainsaw e2e test in `tests/e2e-chainsaw/v1beta2/password-secret-rotation/redis-replication/`: it stands up a 3-node RedisReplication, rotates the password Secret, waits for the checksum annotation on the StatefulSet's pod template to change and the rollout to complete, then verifies the pod accepts the new password and rejects the old one. It's passing in CI (E2E Test check is green).

pujitha24:
Added the chainsaw e2e test as requested and it's still green along with the rest of CI — let me know if there'\''s anything else that would make this easier to review.

PR #1893, opened 2026-09-08
fix(redis): set dial/read/write timeouts on the service Redis client

### LINKED ISSUE #1892: go-redis client in `internal/service/redis` sets no timeouts — an unreachable Redis pod can stall a reconcile worker indefinitely

```shell
redis-operator version: main (bcffc602) / v0.26.0
```

Yes.

Not environment-specific.

**What did you do?**

Have a Redis pod enter a network-blackhole state (e.g. the pod is up but unresponsive, a stuck NetworkPolicy, or a dropped conntrack entry) while role/health reconciliation runs. The heal/check paths connect **directly** to each pod's Redis port:

- `internal/controller/common/redis/heal.go:105-107` — `IsMaster` per pod, per reconcile
- `internal/controller/common/redis/check.go:75-89` — `GetMasterFromReplication`, two connections per pod

The connection factory `createClient()` in `internal/service/redis/client.go:83-96` sets only `Addr`/`Password`/`DB` (+optional `TLSConfig`):

```go
opts := &rediscli.Options{
    Addr:     s.connectionInfo.GetAddress(),
    Password: s.connectionInfo.Password,
    DB:       0,
}
```

The connection attempt fails within a bounded time, the reconcile returns with an error, and the operator retries on the next reconcile.

**What did you see instead?**

go-redis v9 defaults to **no dial/read/write timeout**, so `Dial`, `READ`, or `WRITE` blocks indefinitely. The reconcile worker stays stuck (holding its concurrency slot) for as long as the blackhole persists; with `MAX_CONCURRENT_RECONCILES` exhausted, the whole operator stops making progress.

### Related context

The exec/go-redis path in k8sutils already bounds its connections and documents the rationale (`internal/k8sutils/redis.go:840-842`):

```go
// defaultRedisClientTimeout ... so an unreachable pod cannot stall a reconcile
defaultRedisClientTimeout = 5 * time.Second
```

applied as `DialTimeout/ReadTimeout/WriteTimeout` in `configureRedisClient` (redis.go:791-793). The `internal/service/redis` path (used by the healer/checker and RedisReplication sentinel logic) does not apply an equivalent bound.

Related prior work, both covering only the exec path:

- #1810 — `execCommandTimeout` configuration (pod exec)
- #1843 — skip unreachable pods during role discovery (caller-side mitigation)

Happy to share further traces or details if useful.

trynocoding:
The go-redis client created by `service.createClient()` (`internal/service/redis/client.go`) sets no `DialTimeout`/`ReadTimeout`/`WriteTimeout`. go-redis v9 defaults to no timeout, so when a Redis pod is unreachable in a way that blackholes packets (stuck pod, NetworkPolicy, dropped conntrack), calls like `IsMaster` (invoked per pod per reconcile from `common/redis/heal.go` and `check.go`) block indefinitely and the reconcile worker holds its concurrency slot until the blackhole clears. Enough such reconciles exhaust `MAX_CONCURRENT_RECONCILES` and the operator stops making progress.

The exec/go-redis path in k8sutils already solved this exact problem with `defaultRedisClientTimeout = 5s` (`internal/k8sutils/redis.go:840-842`, applied in `configureRedisClient`). This change mirrors it on the `internal/service/redis` path.

Fixes #1892

**Type of change**

**Checklist**

**Additional Context**

Tests added in `internal/service/redis/client_test.go`:

1. `TestCreateClientSetsTimeouts` — asserts the three timeout options are set (via `client.Options()`).
2. `TestCreateClientUnresponsiveServerReturns` — regression test against a local TCP listener that accepts but never responds: `PING` must return an error within the read timeout (~5s) instead of hanging. Before the fix this call blocks until the 30s test context expires; after the fix it returns in ~5s.

The timeout value (5s) intentionally matches `k8sutils.defaultRedisClientTimeout` so both connection paths behave identically. No configuration knob was added to keep the change minimal; if maintainers prefer this to be configurable (like `execCommandTimeout` in #1810), happy to follow that pattern instead.

### DIFF
--- a/internal/service/redis/client.go
+++ b/internal/service/redis/client.go
@@ -6,12 +6,17 @@ import (
 	"net"
 	"strconv"
 	"strings"
+	"time"
 
 	rediscli "github.com/redis/go-redis/v9"
 )
 
 const (
 	redisRoleMaster = "role:master"
+	// defaultRedisClientTimeout bounds dial/read/write operations so an
+	// unreachable pod cannot stall a reconcile worker. It mirrors the
+	// defaultRedisClientTimeout used by the k8sutils exec-based path.
+	defaultRedisClientTimeout = 5 * time.Second
 )
 
 type ConnectionInfo struct {
@@ -85,9 +90,12 @@ func (s *service) createClient() *rediscli.Client {
 		return nil
 	}
 	opts := &rediscli.Options{
-		Addr:     s.connectionInfo.GetAddress(),
-		Password: s.connectionInfo.Password,
-		DB:       0,
+		Addr:         s.connectionInfo.GetAddress(),
+		Password:     s.connectionInfo.Password,
+		DB:           0,
+		DialTimeout:  defaultRedisClientTimeout,
+		ReadTimeout:  defaultRedisClientTimeout,
+		WriteTimeout: defaultRedisClientTimeout,
 	}
 	if s.connectionInfo.TLSConfig != nil {
 		opts.TLSConfig = s.connectionInfo.TLSConfig
--- a/internal/service/redis/client_test.go
+++ b/internal/service/redis/client_test.go
@@ -1,0 +1,61 @@
+package redis
+
+import (
+	"context"
+	"net"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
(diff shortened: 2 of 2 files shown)


shubham-cmyk:
@trynocoding can you fix the CI lint

PR #1895, opened 2026-09-08
fix(k8sutils): avoid infinite loop in ExecuteRedisReplicationCommand

### LINKED ISSUE #1894: RedisCluster reconcile deadlocks: `ExecuteRedisReplicationCommand` infinite loop when follower count < leader count

```shell
redis-operator version: main (bcffc602) / v0.26.0
```

Yes — by code-path inspection and an in-package reproduction (see Verification below). The hang is reachable with any CR where `0 < follower < leader`; nothing (CRD validation, webhook, or caller guard) currently constrains that configuration.

Not environment-specific — this is a logic bug in the reconcile path.

**What did you do?**

1. Create a `RedisCluster` with `size: 3` (leaders) and `followers: 1` — a valid-looking spec:

```yaml
apiVersion: redis.redis.opstreelabs.in/v1beta2
kind: RedisCluster
metadata:
  name: redis-cluster
spec:
  size: 3            # leaders
  followerSize: 1    # <-- follower < leader
```

2. Wait for leaders to bootstrap and the controller to reach the follower-adding branch of the reconcile (`internal/controller/rediscluster/rediscluster_controller.go:327`).

The reconcile terminates — either replication gets configured, or the failure is surfaced as an error/event.

**What did you see instead?**

The reconcile worker blocks **forever** inside `k8sutils.ExecuteRedisReplicationCommand`, with no error logged. With the default `MAX_CONCURRENT_RECONCILES=3`, a handful of such CRs permanently occupy all workers and the operator stops reconciling anything.

A second, related hang mode affects specs where `follower >= leader` (see Root cause, defect 2): any window where a follower pod is unreachable (rolling update, restart) stalls that reconcile until the pod answers again — while issuing repeated pod GETs against the API server each iteration.

### Root cause (two defects in one loop)

**Defect 1 — integer division + loop with no post statement (`internal/k8sutils/redis.go:487-532`):**

```go
followerPerLeader := followerCounts / leaderCounts   // redis.go:491 → 1/3 = 0
...
for followerIdx := 0; followerIdx <= int(followerCounts)-1; {   // redis.go:500: no post statement
    for i := 0; i < int(followerPerLeader) && followerIdx <= int(followerCounts)-1; i++ {
        ...
        followerIdx++    // redis.go:530: the ONLY increment, inside the inner loop
    }
}
```

When `follower < leader`, `followerPerLeader` truncates to `0`, the inner loop condition is false on the first check, `followerIdx` never increments, and the outer loop (which has no post statement) spins forever. Note that `0 < follower < leader` is a legitimate Redis Cluster topology (fewer replicas than masters), and the CRD exposes leader/follower replicas as independent fields — so such specs are straightforward to write today.

**Defect 2 — `continue` skips the increment:**

At `redis.go:519`, a failed `Ping` executes `continue`, which skips `followerIdx++` at `redis.go:530` (the inner loop still advances `i`). When `i` is exhausted the inner loop exits with `followerIdx` unchanged, the outer loop restarts `i` at 0, and the same pod is retried indefinitely within a single reconcile — for as long as the pod stays unreachable.

### Verification

I reproduced both hang modes with a minimal in-package reproduction (fake clientset, no pods, unreachable Redis port so every dial fails fast):

- `leaders=3, followers=1`: the call never returns; the goroutine is `[runnable]` spinning at `redis.go:500` (the outer loop condition — the inner loop body never executes).
- `leaders=3, followers=3` with unreachable pods: the call never returns; the stack alternates between `redis.go:515` (Ping, in go-redis retry sleep) and the outer loop restart at `redis.go:501`, confirming the ping-fail `continue` skips `followerIdx++` and the same pod is retried forever.

Happy to share the reproduction test or gather additional details if useful.

### Related

- #1733 — a different bug in the same function (`ExecuteRedisReplicationCommand`), suggesting this function deserves a consolidated look.
- #310 (2022, closed) — earlier report of incorrect follower upscaling behavior.

trynocoding:
`ExecuteRedisReplicationCommand` (RedisCluster reconcile, follower-adding branch) can spin forever in two ways:

1. **followers < leaders**: `followerPerLeader := followerCounts / leaderCounts` truncates to `0`, the inner loop never runs, and the outer loop (no post statement) spins forever — the reconcile worker blocks permanently, and a few such CRs exhaust `MAX_CONCURRENT_RECONCILES`.
2. **Unreachable follower**: a failed `Ping` executes `continue`, skipping the `followerIdx++` inside the inner loop body — the same pod is retried indefinitely within a single reconcile.

Fixes #1894

**What changed**

- Flatten the nested loop into a single loop over `followerIdx`; the loop head is the only place the index advances, so no failure path can stall progress.
- Drop `followerPerLeader`: leader pairing is round-robin (`followerIdx % leaderCounts`), so the value never carried meaning beyond an inner batch size — and a batch size of 0 was fatal.
- Return an error on ping / cluster-nodes failures instead of logging and spinning; the caller now propagates it so controller-runtime owns retries with backoff (matching the surrounding `ClusterHasEmptyMasters` error handling and `ExecuteFailoverOperation`'s error-returning signature).
- Keep the exported call site simple: `ExecuteRedisReplicationCommand` delegates to an injectable core, following the existing `podExecFunc` pattern used by `executeSingleLeaderAddSlots`.

**Tests**

`internal/k8sutils/redis_replication_test.go` (new) drives the function against a minimal RESP fake (HELLO / PING / CLUSTER NODES, RESP3 handshake included) with a fake clientset and a recorded exec func:

1. all followers added round-robin (2 leaders / 3 followers, pairing asserted per execution)
2. followers=0 → no-op
3. followers < leaders (1/3) — the case that used to hang forever
4. followers already present in `CLUSTER NODES` are skipped before ping
5. unreachable follower → error naming the pod, no later followers attempted
6. failed `CLUSTER NODES` lookup → error, nothing executed
7. non-PONG ping reply → skip execution, no error

**Type of change**

**Checklist**

**Additional Context**

Verified the original hang empirically before the fix (watchdog + goroutine stack dumps showed the worker spinning at the outer loop condition / ping retry), and the same probes return instantly after the fix. `ExecuteRedisReplicationCommand` previously returned nothing, so the reconcile could not observe failures; the signature change is what lets the operator retry with backoff instead of blocking a worker.

Happy to adjust the error granularity (e.g. collect all failures with `errors.Join`) or anything else maintainers prefer.

### DIFF
--- a/internal/controller/rediscluster/rediscluster_controller.go
+++ b/internal/controller/rediscluster/rediscluster_controller.go
@@ -326,7 +326,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 			if followerReplicas > 0 {
 				logger.Info("All leader are part of the cluster, adding follower/replicas", "Leaders.Count", leaderCount, "Instance.Size", leaderReplicas, "Follower.Replicas", followerReplicas)
-				k8sutils.ExecuteRedisReplicationCommand(ctx, r.K8sClient, instance)
+				if err := k8sutils.ExecuteRedisReplicationCommand(ctx, r.K8sClient, instance); err != nil {
+					return ctrl.Result{}, err
+				}
 			} else {
 				logger.Info("no follower/replicas configured, skipping replication configuration", "Leaders.Count", leaderCount, "Leader.Size", leaderReplicas, "Follower.Replicas", followerReplicas)
 			}
--- a/internal/k8sutils/redis.go
+++ b/internal/k8sutils/redis.go
@@ -116,10 +116,12 @@ func getEndpoint(ctx context.Context, client kubernetes.Interface, cr *rcvb2.Red
 	return host + ":" + strconv.Itoa(port)
 }
 
-// podExecFunc matches executeCommand's signature; it is injected into
-// executeSingleLeaderAddSlots so the command assembly and batching logic
-// can be unit tested without a live pod exec.
-type podExecFunc func(ctx context.Context, client kubernetes.Interface, cr *rcvb2.RedisCluster, cmd []string, podName string)
+// podExecFunc matches executeCommandE's signature; it is injected into
+// executeSingleLeaderAddSlots and executeRedisReplicationCommand so the
+// command assembly and batching logic can be unit tested without a live pod
+// exec. It returns the exec error so callers that reconcile with retries can
+// propagate failures instead of only logging them.
+type podExecFunc func(ctx context.Context, client kubernetes.Interface, cr *rcvb2.RedisCluster, cmd []string, podName string) error
 
 // checkRedisCLIAuthInEnv returns true if we can use the pod's REDISCLI_AUTH variable instead of sending redis-cli -a <password>.
 // It checks only variables specified via env[].valueFrom since this is what the operator sets; it does not look at envFrom.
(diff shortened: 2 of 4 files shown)


shubham-cmyk:
@trynocoding fix the CI lint

PR #1905, opened 2026-09-12
fix(cluster): re-apply dynamicConfig on every reconcile once Ready

### LINKED ISSUE #1757: Dynamic config changes in `spec.redisConfig.dynamicConfig` are not applied to running Redis cluster
### Environment
- **Redis Operator Version**: v0.23.0
- **Kubernetes Version**: [Your K8s version]
- **Deployment Method**: kubectl apply with custom CR YAML

### Problem Description

I created a RedisCluster CR with dynamic configuration using `spec.redisConfig.dynamicConfig`. The initial configuration was applied successfully, but subsequent updates to the `dynamicConfig` field are not being propagated to the running Redis instances.

### Steps to Reproduce

1. Create a RedisCluster CR with initial dynamic config:

```yaml
apiVersion: redis.redis.opstreelabs.in/v1beta2
kind: RedisCluster
metadata:
  name: my3
spec:
  clusterSize: 3
  kubernetesConfig:
    image: "quay.io/opstree/redis:v7.0.15"
  redisConfig:
    dynamicConfig:
      - "maxmemory-samples 4"
  # ... other configurations
```

2. Apply the CR:
```bash
kubectl apply -f my3.yaml
```

3. Verify the configuration is applied:
```bash
kubectl exec -it my3-leader-0 -- redis-cli CONFIG GET maxmemory-samples
# Output shows: 4 ✓ (correct)
```

4. Update the dynamic config in the CR:
```yaml
spec:
  redisConfig:
    dynamicConfig:
      - "maxmemory-samples 11"  # Changed from 4 to 11
```

5. Re-apply the updated CR:
```bash
kubectl apply -f my3.yaml
```

6. Check the configuration again (immediately and after waiting 10+ minutes):
```bash
kubectl exec -it my3-leader-0 -- redis-cli CONFIG GET maxmemory-samples
# Output still shows: 4 ✗ (not updated)
```

### Expected Behavior

When I update `spec.redisConfig.dynamicConfig` in the CR and re-apply it, the Redis operator should:
1. Detect the configuration change
2. Apply the new dynamic configuration to all Redis instances (both leaders and followers)
3. The `CONFIG SET` command should be executed on each pod

### Actual Behavior

The dynamic configuration changes are **not applied** to running Redis instances. The configuration remains at the initial value even after:
- Waiting for multiple reconciliation cycles (10+ minutes)
- Re-applying the CR multiple times
- Verifying that the CR in Kubernetes reflects the updated values

### Investigation

After reviewing the source code (`internal/controller/rediscluster/rediscluster_controller.go`), I found that `SetRedisClusterDynamicConfig()` is only called when the cluster transitions to the `Ready` state for the first time:

```go
// Line 315-323
if instance.Status.ReadyLeaderReplicas == leaderReplicas && 
   instance.Status.ReadyFollowerReplicas == followerReplicas && 
   instance.Status.State != rcvb2.RedisClusterReady {
    
    // Apply dynamic config to all Redis instances in the cluster
    if err = k8sutils.SetRedisClusterDynamicConfig(ctx, r.K8sClient, instance); err != nil {
        logger.Error(err, "Failed to set dynamic config")
        return intctrlutil.RequeueE(ctx, err, "failed to set dynamic config")
    }
    // ...
}
```

The condition `instance.Status.State != rcvb2.RedisClusterReady` prevents the dynamic config from being re-applied once the cluster is already in the `Ready` state.

### Workarounds

Currently, I have to use one of these workarounds to apply config changes:

1. **Manual CONFIG SET** on each pod:
   ```bash
   kubectl exec -it my3-leader-0 -- redis-cli CONFIG SET maxmemory-samples 11
   # Repeat for all pods...
   ```

2. **Delete pods** to trigger recreation:
   ```bash
   kubectl delete pod -l cluster=my3 --grace-period=0 --force
   ```

3. **Modify other CR fields** to trigger StatefulSet update

### Questions

1. Is this the intended behavior? Shouldn't `dynamicConfig` changes be continuously reconciled?

2. If this is by design, what is the recommended approach for updating dynamic configuration on a running cluster?

3. Would it make sense to add logic that detects changes in `dynamicConfig` and applies them during every reconciliation cycle, regardless of the cluster state?

### Additional Context

According to the documentation ([RedisCluster Configuration](https://ot-container-kit.github.io/redis-operator/docs/configuration/rediscluster/)), dynamic configurations should be applied to all Redis instances, and the operator "ensures all accessible instances receive the configuration." However, it doesn't mention that this only happens during initial cluster bootstrap.

This seems like it could be either:
- A bug that needs fixing
- A missing feature that should be implemented
- Or the documentation needs clarification about this limitation

pujitha24:
**Description**

`spec.redisConfig.dynamicConfig` changes on a `RedisCluster` were only applied once, at the moment the cluster first transitioned into the `Ready` state. Once `instance.Status.State == RedisClusterReady`, the `instance.Status.State != rcvb2.RedisClusterReady` guard around the `SetRedisClusterDynamicConfig` call permanently skipped it on every later reconcile, so a subsequent edit to `dynamicConfig` was never pushed to the running leader/follower pods, even after waiting many reconcile cycles or re-applying the CR.

This moves the `SetRedisClusterDynamicConfig` call out of the "just became Ready" block and into a separate block gated only on `instance.Status.State == rcvb2.RedisClusterReady`, so it now runs on every reconcile once the cluster is ready (the controller already requeues every 10s while running), matching how the Redis standalone and RedisReplication controllers already apply their dynamic config on every reconcile rather than only once.

Also refactored `SetRedisClusterDynamicConfig` in `internal/k8sutils/redis.go` to delegate to an injectable-client `setRedisClusterDynamicConfig` helper, mirroring the existing `setRedisReplicationDynamicConfig` / `setRedisStandaloneDynamicConfig` pattern in the same file, so the fix is unit-testable without a live Redis.

Fixes #ISSUE

**Type of change**

**Checklist**

**Additional Context**

Validation:
- `go build ./...` and `go vet ./...` pass.
- `golangci-lint run ./...` passes with 0 issues.
- `go test ./...` passes (including the pre-existing `internal/controller/rediscluster` envtest/ginkgo suite, run with `KUBEBUILDER_ASSETS` from `setup-envtest use 1.31.0 -p path`).
- Added `TestSetRedisClusterDynamicConfig` in `internal/k8sutils/redis_dynamic_config_test.go`, including a subtest that calls the helper twice with a changed config value (simulating two reconciles with a spec edit in between) and asserts `CONFIG SET` is issued with the new value both times via `redismock` expectations.

Disclosure: there is no live-Redis / end-to-end reproduction of this in the test suite — the existing `internal/controller/rediscluster` envtest suite has no real Redis pods, and `applyDynamicConfig` intentionally no-ops (returns no error) when a pod is unreachable, so it cannot fail-then-pass at that layer. The fix is proven at the `k8sutils` unit level (repeated-invocation test above) plus direct inspection of the controller gating change. One side effect worth noting: on the very reconcile where the cluster first transitions to Ready, the new `dynamicConfig` application is skipped because the transition writes the new state via `updateStatus` rather than mutating the local `instance` in place, so dynamic config on a brand-new cluster is first applied on the following reconcile (~10s later) rather than synchronously — this only affects the very first Ready transition, not the reported bug (repeated updates after Ready).

Fixes #1757

### DIFF
--- a/internal/controller/rediscluster/rediscluster_controller.go
+++ b/internal/controller/rediscluster/rediscluster_controller.go
@@ -412,11 +412,6 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		monitoring.RedisClusterHealthy.WithLabelValues(instance.Namespace, instance.Name).Set(0)
 		if k8sutils.RedisClusterStatusHealth(ctx, r.K8sClient, instance) {
 			monitoring.RedisClusterHealthy.WithLabelValues(instance.Namespace, instance.Name).Set(1)
-			// Apply dynamic config to all Redis instances in the cluster
-			if err = k8sutils.SetRedisClusterDynamicConfig(ctx, r.K8sClient, instance); err != nil {
-				logger.Error(err, "Failed to set dynamic config")
-				return intctrlutil.RequeueE(ctx, err, "failed to set dynamic config")
-			}
 
 			requeue, err := r.updateStatus(ctx, instance, rcvb2.RedisClusterStatus{
 				State:                 rcvb2.RedisClusterReady,
@@ -430,6 +425,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 			if requeue {
 				return intctrlutil.Requeue()
 			}
+		}
+	}
+
+	// Apply dynamic config to all Redis instances in the cluster on every reconcile
+	// once the cluster is ready, so that changes to spec.redisConfig.dynamicConfig
+	// made after the initial bootstrap are picked up instead of only being applied
+	// once during the transition into the Ready state.
+	if instance.Status.State == rcvb2.RedisClusterReady {
+		if err = k8sutils.SetRedisClusterDynamicConfig(ctx, r.K8sClient, instance); err != nil {
+			logger.Error(err, "Failed to set dynamic config")
+			return intctrlutil.RequeueE(ctx, err, "failed to set dynamic config")
 		}
 	}
 
--- a/internal/controller/rediscluster/rediscluster_controller.go
+++ b/internal/controller/rediscluster/rediscluster_controller.go
@@ -413,6 +413,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		if k8sutils.RedisClusterStatusHealth(ctx, r.K8sClient, instance) {
 			monitoring.RedisClusterHealthy.WithLabelValues(instance.Namespace, instance.Name).Set(1)
 
+			// Apply dynamic config before persisting the Ready status, so that a
+			// failing CONFIG SET (e.g. an invalid setting) prevents the cluster
+			// from being reported Ready with an unapplied configuration.
+			if err = k8sutils.SetRedisClusterDynamicConfig(ctx, r.K8sClient, instance); err != nil {
+				logger.Error(err, "Failed to set dynamic config")
+				return intctrlutil.RequeueE(ctx, err, "failed to set dynamic config")
+			}
+
 			requeue, err := r.updateStatus(ctx, instance, rcvb2.RedisClusterStatus{
 				State:                 rcvb2.RedisClusterReady,
 				Reason:                rcvb2.ReadyClusterReason,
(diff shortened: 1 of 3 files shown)


shubham-cmyk:
@pujitha24 can you look at the above review comment

